PR #1, opened 2017-11-02
Add initial implementation

### LINKED ISSUE #2: Add initial implementation

lhorie:
Resolves #2


.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  extends: [require.resolve('eslint-config-fusion')],
+  rules: {
+    'no-unused-vars': [
+      'error',
+      {vars: 'all', args: 'none', varsIgnorePattern: '^h$'},
```
rtsao: Is this mean to be here? I assume this has to do with JSX?

lhorie: I think it might be an old config. It seems to be a bad copy-paste from some preact thing

lhorie:
!merge

README.md
```diff
@@ -0,0 +1,49 @@
...
+const Hello = () => <div>Hello</div>
+
+export default function() {
+  return new App(<Hello />);
+}
+```
+
+---
+
+### API
```
rtsao: Docs for HOC?

ganemone: Honestly I think we should maybe not document it at all. I'm not convinced we want to expose this interface

lhorie: my 2 cents: document, and if we change it, do it formally via a changelog/version bump

PR #11, opened 2017-11-29
Add support for overriding render function

### LINKED ISSUE #12: Add support for overriding the render function
It can be useful to overwrite the render function for testing. For example, you may want to swap out the renderer enzyme shallow rendering.

ganemone:
Fixes #12


ganemone:
!merge

PR #37, opened 2018-01-19
Pass `ctx` as prop to provider component

### LINKED ISSUE #36: Pass `ctx` as prop to provider component

ganemone:
Fixes #36

### DIFF
--- a/src/__tests__/hoc.node.js
+++ b/src/__tests__/hoc.node.js
@@ -13,6 +13,7 @@ tape('hoc', async t => {
   function TestComponent(props) {
     didRender = true;
     t.deepLooseEqual(props.test, testProvides);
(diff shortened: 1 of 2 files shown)


src/__tests__/hoc.node.js
```diff
@@ -13,6 +13,7 @@ tape('hoc', async t => {
   function TestComponent(props) {
     didRender = true;
     t.deepLooseEqual(props.test, testProvides);
+    t.notok(props.ctx, 'does not pass ctx through by default');
```
AlexMSmithCA: I read this as "no tok" and was confused.

ganemone: lol

PR #60, opened 2018-01-31
Update provider plugin to use proptypes.any rather than proptypes.object

### LINKED ISSUE #59: Update provider plugin to use proptypes.any rather than proptypes.object
We shouldn't make assumptions about the types of thing being provided.

ganemone:
Fixes #59


ganemone:
!merge

PR #102, opened 2018-04-04
Make Provider middleware idempotent

KevinGrandon:
Simulation tests will currently call middleware functions several times in a row. This middleware fails due to changing originalMiddleware, and is not idempotent. This should allow us to call this middleware several times with the same result.

The issue is seen in the simulation tests added here: https://github.com/fusionjs/fusion-cli/pull/302

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -9,11 +9,12 @@ import Provider from './provider';
 
 export default {
   create: (name, plugin, provider) => {
-    let originalMiddleware = plugin.middleware;
+    let originalMiddlewareGetter = plugin.middleware;
+    let originalMiddleware = null;
     const ProviderComponent = provider || Provider.create(name);
     plugin.middleware = (deps, provides) => {
-      if (originalMiddleware) {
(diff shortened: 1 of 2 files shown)


src/plugin.js
```diff
@@ -9,11 +9,12 @@ import Provider from './provider';
...
 export default {
   create: (name, plugin, provider) => {
-    let originalMiddleware = plugin.middleware;
+    let originalMiddlewareGetter = plugin.middleware;
+    let originalMiddleware = null;
     const ProviderComponent = provider || Provider.create(name);
     plugin.middleware = (deps, provides) => {
-      if (originalMiddleware) {
-        originalMiddleware = originalMiddleware(deps, provides);
+      if (originalMiddlewareGetter && originalMiddleware === null) {
```
ganemone: I think this will cause other problems. For example, what happens if you create two apps and set up different dependencies? The caching here will persist across apps and you would get a plugin with middleware using the wrong set of dependencies.

PR #103, opened 2018-04-04
Update provider plugin to create middleware idempotent

### LINKED ISSUE #104: Issue with simulation tests and fusion-react provider plugins
Simulation tests will currently call middleware functions several times in a row. This middleware fails due to changing originalMiddleware, and is not idempotent. This should allow us to call this middleware several times with the same result.

The issue is seen in the simulation tests added here: fusionjs/fusion-cli#302

ganemone:
Fixes #104


ganemone:
!merge

PR #118, opened 2018-05-16
Improve Flow specificity

AlexMSmithCA:


src/plugin.js
```diff
@@ -8,16 +8,33 @@
...
+  {__plugin__: boolean}
+>;
+
 export default {
-  create: (
+  create: <TDeps: *, TService: *>(
     name: string,
-    plugin: FusionPlugin<*, *>,
+    // eslint-disable-next-line
+    plugin: FusionPluginNoHidden<TDeps, TService> | FusionPlugin<TDeps, TService>,
```
AlexMSmithCA: We allow folks to pass in non-plugins that contain `middleware`, `provides`, and `deps`.  Technically these are not `FusionPlugin`s.

Alternatively, it might be nicer to just use a custom type here:
```
type PartialPlugin<Deps, Service> = {
  deps?: Deps,
  provides?: (Deps: $ObjMap<Deps & {}, ExtractReturnType>) => Service,
  middleware?: (
    Deps: $ObjMap<Deps & {}, ExtractReturnType>,
    Service: Service
  ) => Middleware
};
...
export default { 
     ...
     plugin: PartialPlugin<TDeps, TService> | FusionPlugin<TDeps, TService>
     ...
};
```

This avoids the shenanigans of exposing `__plugin__` outside of `fusion-core`.  But also means we have lose the explicit dependency between these two types.  Thoughts?

package.json
```diff
@@ -23,25 +23,25 @@
...
-    "nyc": "^11.6.0",
+    "nyc": "^11.8.0",
     "prettier": "^1.12.1",
     "prop-types": "^15.6.1",
     "react": "^16.3.2",
     "react-dom": "^16.3.2",
-    "tape-cup": "4.7.1",
-    "unitest": "2.1.1"
+    "tape-cup": "^4.7.1",
+    "unitest": "^1.1.1"
```
KevinGrandon: Looks like these deps got downgraded somehow?

AlexMSmithCA: [v1.1.1 appears to be the latest published version](https://www.npmjs.com/package/unitest).

src/plugin.js
KevinGrandon: Where do we use non-plugins today? Should we enforce that they are Fusion plugins?

package.json
KevinGrandon: @rtsao  - Something we should probably look into.

AlexMSmithCA:
Note that this should land *after* https://github.com/fusionjs/fusion-core/pull/222

src/plugin.js
AlexMSmithCA: Example usage in [src/__tests__/plugin.js](https://github.com/fusionjs/fusion-react/blob/master/src/__tests__/plugin.js#L42).  Without this, they would have to call `createPlugin` and pass in the returned value.

PR #142, opened 2018-08-03
Update README.md

maheshbabu:
`export 'split' was not found in 'fusion-react'` 

Fix for the above error due to wrong import. split seems to be available from 'fusion-react-async'

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ Consider using [`fusion-react`](https://github.com/fusionjs/fusion-react) instea
 #### split
 
 ```js
-import {split} from "fusion-react";
+import {split} from "fusion-react-async";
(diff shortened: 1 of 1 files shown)


rtsao:
@maheshbabu fusion-react-async was merged into fusion-react in `fusion-react@1.1.0`, so possibly you're still on 1.0.x?

KevinGrandon:
Thanks for the info Ryan. @maheshbabu If you used `yarn create fusion-app`, we were previously using an earlier version of fusion-react. Upgrading dependencies here: https://github.com/fusionjs/create-fusion-app/pull/29

Please comment or re-open if you're still seeing this after upgrading.

PR #143, opened 2018-08-03
Call componentWillUnmount in the browser

ganemone:
Some components in the browser set up long lived listeners, connections, etc, that need to be cleaned up when the component is unmounted. For example, react-router will listen on the history object. We should call the componentWillUnmount to ensure that anything created during the prepare cycle is cleaned up.


ganemone:
!merge

PR #146, opened 2018-08-07
Add React.forwardRef support in async prepare

rtsao:
Thanks to @puppybits for initial implementation in https://github.com/fusionjs/fusion-react-async/pull/119


rtsao:
!merge

PR #149, opened 2018-08-14
Support using UNSAFE_componentWillMount

KevinGrandon:
* Upgrades dependencies to latest.
* Rename and support using UNSAFE_componentWillMount as React now suggest using this as a stopgap.


KevinGrandon:
!merge

mlmorg:
Might be good to link to https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#gradual-migration-path in the description?

PR #153, opened 2018-08-22
Add support for getDerivedStateFromProps 

### LINKED ISSUE #152: Missing support for getDerivedStateFromProps
### Description

Recently many libraries have started breaking with Fusion.js due to `getDerivedStateFromProps` usage not being called by prepare.

### Current behavior

* Clone and setup this repository: https://github.com/IcculusC/fusion-material-ui-bug-lab
* Notice that an error occurs:

image

* The following server warning also appears: `Warning: Can't call setState on a component that is not yet mounted. This is a no-op, but it might indicate a bug in your application. Instead, assign to `this.state` directly or define a `state = {};` class property with the desired state in the TransitionGroup component.`

### Expected behavior

The application should work as expected.

### Your environment

* fusion-react version: 1.2.0

KevinGrandon:
In React 16 a new function has been added, getDerivedStateFromProps . This function is used as a hook for deriving state, and is replacing componentWillMount. Many libraries have started leveraging this hook and will throw an error if render is called before this is run. Example: https://github.com/reactjs/react-transition-group/blob/c9b7bbdae2575bfe8c72b564b4c4e5ac9ecb5ebb/src/TransitionGroup.js#L109-L119

Fixes #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,6 @@ const hoc = prepared(sideEffect, opts);
   - `componentDidMount: boolean` - Optional. Defaults to `true`. On the browser, `sideEffect` is called when the component is mounted.
   - [TO BE DEPRECATED] `componentWillReceiveProps: boolean` - Optional. Defaults to `false`. On the browser, `sideEffect` is called again whenever the component receive props.
   - `componentDidUpdate: boolean` - Optional. Defaults to `false`. On the browser, `sideEffect` is called again right after updating occurs.
-  - `getDerivedStateFromProps: boolean` - Optional. Defaults to `false`. On the browser, `sideEffect` is called when getDerivedStateFromProps is triggered.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -21,19 +21,19 @@ as bundle splitting and `fusion-react` provides tools to do it easily.
...
-  * [Provider](#provider)
-  * [ProviderPlugin](#providerplugin)
-  * [ProvidedHOC](#providedhoc)
-  * [middleware](#middleware)
-  * [split](#split)
-  * [prepare](#prepare)
-  * [prepared](#prepared)
-  * [exclude](#exclude)
-* [Examples](#examples)
+- [Installation](#installation)
```
KevinGrandon: Prettier. 🤷‍♂️

KevinGrandon:
!merge

PR #160, opened 2018-08-31
Revert "Call componentWillUnmount in the browser"

ganemone:
Calling componentWillUnmount during prepare in the browser causes more problems than it solves. Ultimately I think this is something that will only be truly fixed once react suspense is released. Reverting this for now.

This reverts commit bc10467ef0144abe1add8e74dda90ebdf853e2b4.


ganemone:
!merge

PR #163, opened 2018-09-12
Add Dependency-Injection Higher-Order Component

shannonmoeller:
RFC: https://github.com/fusionjs/rfcs/pull/11
Depends on: https://github.com/fusionjs/fusion-core/pull/290


src/injector.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export function registerInjector(app: FusionApp) {
+  function inject(deps: Dependencies): Services {
+    return getServices(app, deps);
+  }
+
+  const injectorPlugin = createPlugin({
+    middleware: () => (ctx, next) => {
+      ctx.element = (
```
ganemone: This should check for `ctx.element` before wrapping it. `ctx.element` is only present on server side renders

shannonmoeller:
@ganemone I've added support for the legacy api if the new context api isn't available. This is easily removed once support for older versions of react is dropped.

shannonmoeller:
That coverage report is wrong.

src/injector.js
```diff
@@ -0,0 +1,129 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import * as React from 'react';
+import FusionApp, {createPlugin} from 'fusion-core';
+import PropTypes from 'prop-types';
+
+type Dependencies = {[string]: any};
```
ganemone: I think we can refine these types a bit, but this is ok for now

ganemone:
!merge

rasmusprentow:
Nice! This actually solves this issue https://github.com/fusionjs/fusion-react/issues/106.

Good Job!

PR #165, opened 2018-09-21
Show a better error if `prepare` encounters invalid React element

lhorie:
If `prepare` receives something that is not a well-formed React element, it barfs with an unhelpful error (type is not a function)

It should instead throw an error about what is most likely to cause the error to happen

### DIFF
--- a/src/async/prepare.js
+++ b/src/async/prepare.js
@@ -15,6 +15,7 @@ import {
 } from 'react-is';
 
 import isReactCompositeComponent from './utils/isReactCompositeComponent';
+import isReactFunctionalComponent from './utils/isReactFunctionalComponent';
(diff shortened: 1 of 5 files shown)


src/async/prepare.js
```diff
@@ -15,6 +15,7 @@ import {
 } from 'react-is';
 
 import isReactCompositeComponent from './utils/isReactCompositeComponent';
+import isReactFunctionalComponent from './utils/isReactFunctionalComponent';
```
KevinGrandon: This is a new case, correct? Would be amazing if we could have a unit test for this.

lhorie: Added tests for both utils

PR #183, opened 2018-10-11
Support portals

### LINKED ISSUE #176: prepare function fails for react portals
We have an element in our app which is an instance of [React Portal](https://reactjs.org/docs/portals.html) and it seems to not have `type` field although the element itself is valid.

I think that `prepare` function assumes every React Element has a `type` function.

```js
 Uncaught TypeError: type is not a function
    at prepareElement (prepare.js:54)
    at _prepare (prepare.js:71)
    at prepare.js:74
    at Array.map (<anonymous>)
    at prepare.js:72
```

Workaround was to [exclude](https://github.com/fusionjs/fusion-react#exclude) this el.

lhorie:
Fixed #176

Portals are not technically supported server-side, but `prepare` should still not throw if they are used client-side

### DIFF
--- a/src/async/__tests__/index.browser.js
+++ b/src/async/__tests__/index.browser.js
@@ -1,0 +1,14 @@
+/** Copyright (c) 2018 Uber Technologies, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+/* eslint-env node */
+import './__browser__/prepare-render.browser.js';
+
+process.on('unhandledRejection', e => {
+  throw e;
+});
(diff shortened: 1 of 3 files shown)


src/async/__tests__/index.browser.js
```diff
@@ -0,0 +1,14 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+/* eslint-env node */
+import './__browser__/prepare-render.browser.js';
+
+process.on('unhandledRejection', e => {
```
KevinGrandon: Weird, why do we need this file? I don't think we have it for other tests, right?

lhorie: oops, bad copy paste

PR #184, opened 2018-10-11
Run argument validation early

lhorie:
Rationale: when erroneously calling `new App(SomeComponent)` instead of `new App(<SomeComponent />)`, an error gets thrown from either `prepare` or a random `React.Children.only` call if such a thing exists in the Provider tree setup by the DI system. The error is extremely difficult to debug and does not provide any hint towards the source of the problem.

This problem has appeared a number of times in internal Uber migrations and wasted a lot of developer time.

This PR adds a validation check in the `new App` call, in order to throw a more debuggable/actionable error in this situation.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -25,6 +25,7 @@ declare var __NODE__: Boolean;
 
 export default class App extends FusionApp {
   constructor(root: React.Element<*>, render: ?(React.Element<*>) => any) {
+    if (!React.isValidElement(root)) throw new Error('Invalid React element');
     const renderer = createPlugin({
       deps: {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -25,6 +25,10 @@ declare var __NODE__: Boolean;
 
 export default class App extends FusionApp {
   constructor(root: React.Element<*>, render: ?(React.Element<*>) => any) {
+    if (!React.isValidElement(root))
+      throw new Error(
+        'Invalid React element. Ensure your root element is a React.Element and not a React.Component'
```
KevinGrandon: nit: `.` at the end?

PR #192, opened 2018-11-07
Upgrade dependencies and createContext typing

KevinGrandon:
Updates code to pass when using flow 0.85


AlexMSmithCA:
I think we should add the `bugfix` label to these Flow fixes to trigger a future patch release.

KevinGrandon:
> Let's add the `bugfix` label to these Flow fixes to trigger a future patch release.

Sounds good, does `dependencies` not trigger a patch? Feels like maybe it should?

KevinGrandon:
!merge

PR #210, opened 2019-01-11
Update dep versions and fix flow errors

lhorie:
Some new flow errors came up with flow 0.90.0


AlexMSmithCA:
Thanks for handling these!

PR #227, opened 2019-02-08
Update async rendering for React 16

rtsao:
**Increased minimum React version**
- `react` and `react-dom` peer dependencies changed to `^16.8.1`

**Improvements to async rendering**
- Support for hooks
- Eliminate client-side prepare step
- Eliminate `__IS_PREPARE__` context, whose only purpose was during client-side prepare.
- Replace custom React tree traversal with `renderToStaticMarkup`

**Flow type improvements**
- `prepared` and `split` now have more precise flow types

### Breaking change with `prepared` HOC

Prepared components now only call the effectFn once, regardless of how many times it is rendered. To restore the old behavior, add unique `effectId` props.

```js

const PreparedComponent = prepared(effectFn)(SomeComponent);

// effectFn called only once
const app1 = (
  <div>
    <PreparedComponent />
    <PreparedComponent />
    <PreparedComponent />
  </div>
);

// effectFn called for each rendered PreparedComponent
const app2 = (
  <div>
    <PreparedComponent effectId="1" />
    <PreparedComponent effectId="2" />
    <PreparedComponent effectId="3" />
  </div>
);
```


rtsao:
!merge

PR #229, opened 2019-02-20
Upgrade flow-bin and fusion-core in lockstep

KevinGrandon:


KevinGrandon:
Landing for now to unblock release verification. Feel free to audit.

PR #237, opened 2019-03-05
Add prepare context

rtsao:
This is needed by the react-router plugin now that client-side prepare is run again

### DIFF
--- a/src/async/prepare.js
+++ b/src/async/prepare.js
@@ -71,6 +71,7 @@ export default function prepare(element: any) {
   class PrepareContextProvider extends React.Component<{}> {
     getChildContext() {
       return {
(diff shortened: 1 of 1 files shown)


src/async/prepare.js
```diff
@@ -80,6 +81,7 @@ export default function prepare(element: any) {
   }
   PrepareContextProvider.childContextTypes = {
     __PREPARE_STATE__: () => {},
+    __IS_PREPARE__: () => {},
```
lhorie: Should this be false?

rtsao: No, this is context types, which should be a prop type function. But rather than use that, a noop function suffices instead of using a real prop type.

PR #242, opened 2019-04-01
Add useService hook and modern Context

micburks:
[RFC](https://github.com/fusionjs/rfcs/pull/15)

- Implement modern Context API for services
- Implement useService hook
- add tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,9 +324,9 @@ function Component() {
 }
 ```
 
-- `token: Token` - Required. The token used to lookup the registered plugin.
-- `children: Service => React.Element<any>` - Required. Render prop that is passed the registered service. Should return the React Element to render.
-- `service: any` - The service provided by the registered plugin.
+- `token: Token<TService>` - Required. The token used to lookup the registered plugin.
+- `children: TService => React.Element<any>` - Required. Render prop that is passed the registered service. Should return the React Element to render.
+- `service: TService` - The service provided by the registered plugin.
 
 This is the same pattern as the `useService` hook. Opt for using the hook. `ServiceConsumer` is provided as a replacement for any legacy Context usage that may exist.
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,9 @@ as bundle splitting and `fusion-react` provides tools to do it easily.
   - [prepare](#prepare)
   - [prepared](#prepared)
   - [exclude](#exclude)
+  - [useService](#useservice)
+  - [ServiceConsumer](#serviceconsumer)
+  - [FusionContext](#fusioncontext)
 - [Examples](#examples)
 
 ---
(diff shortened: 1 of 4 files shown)


src/__tests__/context.node.js
```diff
@@ -0,0 +1,61 @@
...
+import {getSimulator} from 'fusion-test-utils';
+import App from '../index';
+import {serviceContextPlugin, ServiceConsumer, useService} from '../context.js';
+
+test('useService hook', async t => {
+  const TestToken = createToken('test');
+  const TestPlugin = createPlugin({provides: () => 3});
+  let didRender = false;
+  function TestComponent() {
+    const provides = useService<typeof TestToken>(TestToken);
```
ganemone: I think we can actually remove the `typeof` here since the return value is not reachable by exports.

src/context.js
```diff
@@ -0,0 +1,54 @@
+// @flow
+import * as React from 'react';
+import {createPlugin} from 'fusion-core';
+import type FusionApp, {FusionPlugin, Middleware, Token} from 'fusion-core';
+import type {Element} from 'react';
+
+type ExtractReturnType = <V>(() => V) => V;
```
ganemone: I believe this utility might be exported from `fusion-core` however it is also fine to define it here.

ganemone:
Only thing missing is an update to the `readme` with documentation on the `useService` hook and might be good to add a test that documents the behavior when you call `useService` on a token with no value registered. Does it throw an error or return null? I think we probably want it to throw, but not totally sure. Worth some discussion

src/context.js
micburks: It's not exported at the top level in core. Would be useful for me to fix it in `fusion-core`?

src/__tests__/context.node.js
micburks: 👍

src/context.js
AlexMSmithCA: For what it's worth, this is a fairly common pattern in Flow, so it's probably fine to have duplicated.

src/context.js
```diff
@@ -0,0 +1,54 @@
...
+
+type ServiceContextType<T> = {
+  (): $Call<ExtractReturnType, T>,
+};
+
+export const ServiceContext = React.createContext<ServiceContextType<any>>(
+  () => {}
+);
+
+export function useService<T: Token<any>>(token: T): ServiceContextType<T> {
```
AlexMSmithCA: Are we able to avoid the usage of `any` here?  With @ganemone's suggestion above to avoid explicitly typing the generic, I think we can improve this.

Maybe something like this:

```
export function useService<TService>(
  token: Token<TService>
): ServiceContextType<TService> {
  const getService: (
    Token<TService>
  ) => ServiceContextType<TService> = React.useContext(ServiceContext);
  const provides = getService(token);
  return provides;
}
```

We could remove the usage for `ExtractReturnType` entirely then, and `ServiceContextType` would just return `T` directly:

```
type ServiceContextType<TService> = {
   (): TService,
 };
```

micburks:
> Only thing missing is an update to the `readme` with documentation on the `useService` hook and might be good to add a test that documents the behavior when you call `useService` on a token with no value registered. Does it throw an error or return null? I think we probably want it to throw, but not totally sure. Worth some discussion

Right now there is no difference between a token that hasn't been registered, a plugin that has no `provides`, or a plugin that has a `provides` that returns `undefined`; they all return `undefined`. We could throw in all these situations, since it doesn't make sense to use `useService` if you're not expecting anything to be returned, or we could add another method `hasService` to the base app in `fusion-core` so we can have more accurate messaging for errors.

src/context.js
ganemone: Sounds good

src/context.js
micburks: Yeah that makes sense. Now that it's been cleared up that you don't need to pass the type into `useService` this will be easier. With this change, is it necessary to have the `ServiceContextType` at all? It just looks like an identity type now

ganemone: Yea that looks good assuming it works like we think.

AlexMSmithCA: Hm, we may even be able to do away with `ServiceContextType` here as `provides` should resolve to the actual service itself (as opposed to an invocable object that resolves to the service).

src/__tests__/context.node.js
```diff
@@ -0,0 +1,61 @@
...
+import {getSimulator} from 'fusion-test-utils';
+import App from '../index';
+import {serviceContextPlugin, ServiceConsumer, useService} from '../context.js';
+
+test('useService hook', async t => {
+  const TestToken = createToken('test');
+  const TestPlugin = createPlugin({provides: () => 3});
+  let didRender = false;
+  function TestComponent() {
+    const provides = useService(TestToken);
```
AlexMSmithCA: (See note below.  With current types, it looks like this resolves to `ServiceContextType<Token<any | number>>` right now)

AlexMSmithCA:
Cool, changes look great!

README.md
```diff
@@ -283,6 +286,72 @@ const NewComponent = exclude(Component);
...
+
+function Component() {
+  const service = useService(ExampleToken);
+  return (
+    <button onClick={service}>Invoke Service</button>
+  );
+}
+```
+
+- `token: Token` - Required. The token used to lookup the registered plugin.
```
AlexMSmithCA: Nitpick: we could avoid `any` here by using the generic on `Token`.  Something like:

```
`token: Token<TService>` - Required.  The token used to look up the registered plugin that resolves to `TService`.
`service: TService` - ...
```

micburks: Makes sense

README.md
```diff
@@ -283,6 +286,72 @@ const NewComponent = exclude(Component);
...
+This is the same pattern as the `useService` hook. Opt for using the hook. `ServiceConsumer` is provided as a replacement for any legacy Context usage that may exist.
+
+#### FusionContext
+
+```js
+import {useContext} from 'react';
+import {FusionContext} from 'fusion-react';
+
+function Component() {
+  const ctx = useContext(FusionContext);
```
micburks: I've been unable to find any client-side plugins that use this pattern. Is this even useful to have in the browser?

ganemone: we might need to do some refactoring, but I definitely think it will be useful. For one example, web-rpc will definitely benefit. Also potentially apollo client

src/context.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+import * as React from 'react';
+import {createPlugin} from 'fusion-core';
+import type FusionApp, {FusionPlugin, Middleware} from 'fusion-core';
+import type {Element} from 'react';
+
+export const FusionContext = React.createContext<any>({});
+export const ServiceContext = React.createContext<any>(() => {});
+
+type ReturnsType<T> = () => T;
```
micburks: Needed a mechanism to support `Token.optional` since flow considers it different from `Token`. This technically allows someone to pass in something that isn't a token. The usage of this function is so straightforward that I don't think this is a huge issue.

micburks:
!merge

PR #249, opened 2019-05-03
Pass all instrumented metadata into splitComponentLoaders

micburks:
With  fusionjs/fusion-cli#769

Pass `__I18N_KEYS` from instrumented promise into `splitComponentLoaders`. The only consumer of `splitComponentLoaders` right now is `fusion-plugin-i18n`, but if the need arises in the future this is simpler to expand upon if all metadata is passed in together.


micburks:
Yes it's breaking technically, and yes we only use it in the i18n plugin. I really only made it breaking to come up with a more extensible API, but that is assuming `splitComponentLoaders` stays around when Suspense comes or if webpack was replaced. It can be avoided if we just pass the keys in as a second parameter, unless you consider a function signature change to be breaking because of flow.

Extra properties on the array is a bit goofy, but at the same time this is a pretty insignificant (and probably temporary) feature so I'm okay if that's the decision.

PR #378, opened 2019-05-13
Update registry to npmjs.org in .yarnrc

KevinGrandon:
I tried using the yarn registry, but was running into package not found errors.


KevinGrandon:
Merging with override button until we get CI passing.

PR #380, opened 2019-05-13
Lock istanbul-api version to get around bug

micburks:
`istanbul-reports` introduced a bug. Locking version of `instanbul-api` to get around that


micburks:
!merge

PR #381, opened 2019-05-13
Update usage of deprecated API

micburks:
`compiler.plugin('make')` => `compiler.hooks.make`

This was outputting a deprecation notice on every build


micburks:
!merge

PR #383, opened 2019-05-13
Add ability to configure transpilation and bundling with experimentalCompileTest

ganemone:
This adds the `experimentalCompileTest` option to .fusionrc.js. It remains 
undocumented as of now. We will first test in a prerelease to make sure
it fits our needs. Once we are ready to release it outside of experimental
mode, we will add documentation.

### DIFF
--- a/fusion-cli/build/get-webpack-config.js
+++ b/fusion-cli/build/get-webpack-config.js
@@ -205,17 +205,17 @@ function getWebpackConfig(opts /*: WebpackConfigOpts */) {
         if (!JS_EXT_PATTERN.test(modulePath)) {
           return false;
         }
-        const {transpile} = experimentalCompileTest(modulePath, {
-          transpile: 'spec',
-          bundle: 'client',
+        const {transform} = experimentalCompileTest(modulePath, {
+          transform: 'spec',
+          bundle: 'browser-only',
         });
-        if (transpile === 'none') {
+        if (transform === 'none') {
           return false;
-        } else if (transpile === 'all' || transpile === 'spec') {
+        } else if (transform === 'all' || transform === 'spec') {
           return true;
         } else {
           throw new Error(
-            `Unknown transpile value from experimentalCompileTest ${transpile}. Expected 'spec' | 'all' | 'none'`
+            `Unknown transform value from experimentalCompileTest ${transform}. Expected 'spec' | 'all' | 'none'`
           );
         }
       }
@@ -227,17 +227,17 @@ function getWebpackConfig(opts /*: WebpackConfigOpts */) {
       if (!JS_EXT_PATTERN.test(modulePath)) {
         return false;
       }
-      const {transpile} = experimentalCompileTest(modulePath, {
-        transpile: fusionConfig.experimentalCompile ? 'all' : 'spec',
-        bundle: 'client',
+      const {transform} = experimentalCompileTest(modulePath, {
+        transform: fusionConfig.experimentalCompile ? 'all' : 'spec',
+        bundle: 'browser-only',
       });
-      if (transpile === 'none' || transpile === 'spec') {
+      if (transform === 'none' || transform === 'spec') {
         return false;
-      } else if (transpile === 'all') {
+      } else if (transform === 'all') {
         return true;
       } else {
         throw new Error(
-          `Unknown transpile value from experimentalCompileTest ${transpile}. Expected 'spec' | 'all' | 'none'`
+          `Unknown transform value from experimentalCompileTest ${transform}. Expected 'spec' | 'all' | 'none'`
         );
       }
     };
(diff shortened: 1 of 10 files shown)


rtsao:
It seems to me that if we add this, we might want to deprecate `experimentalCompile` and then throw an error if both `experimentalCompile` and `experimentalCompileTest` are both supplied. These are two knobs that kind of control the same thing.

fusion-cli/build/load-fusionrc.js
```diff
@@ -16,10 +16,16 @@ const chalk = require('chalk');
 let loggedNotice = false;
 
 /*::
+
+type CompileResult = {
+  bundle: 'both' | 'client',
+  transpile: 'spec' | 'all' | 'none',
```
rtsao: I personally think `transpile` is a misleading term because this is so strongly associated with ES2017 -> ES5, but this is also about macros/loaders.

```suggestion
  transform: 'all' | 'spec' | 'none',
```

fusion-cli/build/load-fusionrc.js
```diff
@@ -16,10 +16,16 @@ const chalk = require('chalk');
 let loggedNotice = false;
 
 /*::
+
+type CompileResult = {
```
rtsao: Do you think the following would be a valid use case?

A module that should be bundled in both the server and client.
But `transpile: "spec"` in the browser, and `transpile: "none"` in the server.

I'm thinking probably not. In theory, the should be no reason to do `transpile: "spec"` for both, assuming the parse is already happening once. Just wanted to make note of the reasoning for posterity.

fusion-cli/build/load-fusionrc.js
```diff
@@ -16,10 +16,16 @@ const chalk = require('chalk');
...
+  bundle: 'both' | 'client',
```
rtsao: What about:
```suggestion
  bundle: 'universal' | 'browser-only',
```

fusion-cli/build/load-fusionrc.js
ganemone: Seems like a valid use case to me, since the module can have completely different source code for the server and browser bundles via different entry points.

rtsao: Okay, unless I'm missing something, this is impossible to express with the current API. Since it seems the `transforms`/`transpile` option would apply to both server and client?

fusion-cli/build/get-webpack-config.js
```diff
@@ -398,12 +395,29 @@ function getWebpackConfig(opts /*: WebpackConfigOpts */) {
...
-            // do not bundle external packages and those not whitelisted
             const absolutePath = resolveFrom.silent(context, request);
+            // do not bundle external packages and those not whitelisted
             if (absolutePath === null) {
               // if module is missing, skip rewriting to absolute path
               return callback(null, request);
             }
+            if (experimentalCompileTest) {
+              const {bundle} = experimentalCompileTest(absolutePath, {
+                transform: 'none', // default transform value doesn't actually matter here
```
rtsao: Shouldn't the default be based on whether in `node_modules` vs `src/`?

rtsao: Oh, I see this function is potentially called twice for each module. Would this be confusing for users? I wonder if we should store the value and reference it in the babel check.

ganemone: From my understanding, that is done via the regex test on line 397. I thought this code branch was already running on only node_modules. 

However, now that I think about it, I think that behavior might change now that we effectively support bundling some node_modules... we need to change that logic.

ganemone: We could also update the API to include the environment, which would allow different configurations for server vs browser.

ganemone: Also, wouldn't returning 'none' for babel result in babel not parsing that file at all? So it seems like it could be useful for performance to have different config for server vs browser here

rtsao:
```suggestion
                transform: 'spec',
```

For browser-only bundling, the default option should be `transform: spec`, right?

ganemone: Technically this is being called to decide whether to bundle a external dependency or not, so the transpile default for external dependencies is none, since they are not bundled. Also, we don't care about the transform response here so it can technically be anything

rtsao: Oh right, the externals function only gets called for bare imports.

rtsao: It seems slightly odd that we don't use the returned transform value at all (and instead just re-invoke the function inside the babel config setup). I wonder if it should just be two functions then (one for bundle value, the other for transform value). Alternatively, it might make sense to cache/memoize the function so it doesn't get re-invoked multiple time.

PR #385, opened 2019-05-13
Handle non-Error promise rejections

KevinGrandon:
It's possible to reject a non-Error value, so we should handle this and coerce it to an Error when that happens.


KevinGrandon:
!merge

PR #390, opened 2019-05-16
Remove babel polyfill, use core-js and regenerator-runtime directly

micburks:
Removes all deprecation notices from fusion-cli logs


fusion-cli/entries/client-public-path.js
```diff
@@ -4,7 +4,8 @@ declare var __webpack_public_path__: string;
 declare var __webpack_nonce__: string;
 */
 
-import '@babel/polyfill';
+import 'core-js/stable';
```
rtsao: Does this still work with the env preset `useBuiltIns: 'entry'` option? I recall this was hardcoded to `@babel/polyfill` for while.

fusion-cli/entries/client-public-path.js
```diff
@@ -4,7 +4,8 @@ declare var __webpack_public_path__: string;
...
+import 'regenerator-runtime/runtime';
```
rtsao: If we're using standalone regenerator, let's use the non-import side effect option, then we can remove this from our side effect import whitelist

rtsao: Oh wait, nevermind, they removed that in favor of import side effects again

fusion-cli/entries/client-public-path.js
micburks: It looks like `useBuiltins: 'entry'` replaces `@babel/polyfill` with the imports directly from `core-js` (https://babeljs.io/docs/en/babel-preset-env#usebuiltins-entry). I'm not entirely sure why the notices told me to replace the import for polyfill since it seems to do it automatically

micburks:
!merge

PR #391, opened 2019-05-16
Add throwing an error when ServiceContext.Provider wasn't registered

micburks:
This should never happen in fusion-react apps, but will happen in minimal test setups that try to use `useService` or the like.

### DIFF
--- a/fusion-react/src/context.js
+++ b/fusion-react/src/context.js
@@ -11,7 +11,11 @@ import type FusionApp, {FusionPlugin, Middleware} from 'fusion-core';
 import type {Element} from 'react';
 
 export const FusionContext = React.createContext<any>({});
-export const ServiceContext = React.createContext<any>(() => {});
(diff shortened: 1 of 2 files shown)


fusion-react/src/context.js
```diff
@@ -11,7 +11,11 @@ import type FusionApp, {FusionPlugin, Middleware} from 'fusion-core';
 import type {Element} from 'react';
 
 export const FusionContext = React.createContext<any>({});
-export const ServiceContext = React.createContext<any>(() => {});
+export const ServiceContext = React.createContext<any>(() => {
+  throw new Error(
+    '`ServiceContext.Provider` was not found. This occurs if you are attempting to use `ServiceContext` in a non-React fusion application.'
```
AlexMSmithCA: Super minor: we should capitalize Fusion/Fusion.js here.

micburks:
!merge

PR #392, opened 2019-05-16
Remove legacy context from styletron plugin

micburks:
At this point we will stop supporting legacy context for our plugins


micburks:
!merge

PR #393, opened 2019-05-16
Add test for polyfill

micburks:
This isn't very complete, but a good signal that the right polyfills are added based on environment


micburks:
!merge

PR #394, opened 2019-05-16
Add web app manifest plugin

jh3y:
### Description
- Adds new plugin for handling web app manifest files 👍 
- Created with `create-fusion-plugin`
- Basic tests for inserting `<link/>` and returning correct request `body`

Do I need to add `.buildkite` files? Not too sure what is required in order to publish etc.

Look forward to feedback.


lhorie:
Looks good. Can you add some docs?

jh3y:
Thanks @lhorie 👍  Where should I add docs? I've added a `README` for the plugin showing how to use it/install it based off of the other plugins in the repo.

Thanks @KevinGrandon 👍 I'll give that a try!

jh3y:
Hey @KevinGrandon 👋 Could I get re-approval on this one please?

fusion-plugin-web-app-manifest/src/server.js
```diff
@@ -0,0 +1,44 @@
...
+import {WebAppManifestToken} from './tokens';
+import type {DepsType} from './types';
+
+const plugin =
+  __NODE__ &&
+  createPlugin({
+    deps: {
+      manifest: WebAppManifestToken,
+    },
+    provides({manifest}) {
```
ganemone: If the provides is just returning an exact copy of the value registered on `WebAppManifestToken`, we actually don't need the `provides` at all. If someone depends on the manifest, they can depend on the `WebAppManifestToken` directly.

Then we could just change the `middleware` to get the manifest from the dependencies, rather than the self provides.

```js
middleware({manifest}) {
  ...
```

fusion-plugin-web-app-manifest/src/types.js
```diff
@@ -0,0 +1,12 @@
+/** Copyright (c) 2019 Uber Technologies, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+export type DepsType = {};
```
ganemone: This should match the types of the dependencies

```js
export type DepsType = {
  manifest: typeof ManifestToken 
}
```

fusion-plugin-web-app-manifest/src/server.js
```diff
@@ -0,0 +1,44 @@
...
+ * @flow
+ */
+import {createPlugin, html} from 'fusion-core';
+import type {FusionPlugin} from 'fusion-core';
+import {WebAppManifestToken} from './tokens';
+import type {DepsType} from './types';
+
+const plugin =
+  __NODE__ &&
+  createPlugin({
```
ganemone: you can pass the types as arguments here:

```js
createPlugin<DepsType, void>
```

fusion-plugin-web-app-manifest/src/types.js
```diff
@@ -0,0 +1,12 @@
...
+
+export type WebAppManifestType = {
```
ganemone: Is this the full type of the app manifest?

chrisdothtml:
@jh3y looks like eslint and tests for your plugin are failing in CI. Could you fix those?

chrisdothtml:
There was some weird updates in the lockfile that were causing an unrelated plugin to fail in CI. I ran `rush update --full --purge` and it should be working now

jh3y:
Thanks so much @chrisdothtml 👍

PR #395, opened 2019-05-17
Add error message if App init doesn't return the App instance

micburks:
Someone had an issue with this yesterday. Seems pretty insignificant, but a definitely frustrating issue to come across without any special error messaging

### DIFF
--- a/fusion-cli/entries/client-entry.js
+++ b/fusion-cli/entries/client-entry.js
@@ -16,6 +16,9 @@ function reload() {
   const main = require('__FUSION_ENTRY_PATH__'); // eslint-disable-line
   const initialize = main.default || main;
   Promise.resolve(initialize()).then(app => {
+    if (!app) {
+      throw new Error('Application entry point did not return an App');
+    }
     if (window.__ROUTE_PREFIX__) {
       // No-op plugin so token can be registered without any consumers
       // Should not be needed when route prefixing is refactored into a separate plugin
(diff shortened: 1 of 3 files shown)


fusion-cli/entries/client-entry.js
```diff
@@ -16,6 +16,9 @@ function reload() {
   const main = require('__FUSION_ENTRY_PATH__'); // eslint-disable-line
   const initialize = main.default || main;
   Promise.resolve(initialize()).then(app => {
+    if (!app) {
+      throw new Error('Application entry point did not return an App');
```
albertywu: should this check be done inside the `initialize()` function to avoid repeating?

fusion-cli/entries/client-entry.js
```diff
@@ -16,6 +16,9 @@ function reload() {
   const main = require('__FUSION_ENTRY_PATH__'); // eslint-disable-line
   const initialize = main.default || main;
   Promise.resolve(initialize()).then(app => {
+    if (!app) {
```
KevinGrandon: Should we check if this is an instance of BaseApp from fusion-core?

fusion-cli/entries/client-entry.js
micburks: Unfortunately we can't. `initialize` is the default export from `src/main.js` in the consumer application. This check was just to make sure that function always returns a valid application

fusion-cli/entries/client-entry.js
micburks: That works!

micburks:
!merge

PR #397, opened 2019-05-17
Remove legacy context, refactor HOC in fusion-plugin-rpc-redux-react

micburks:
`fusion-plugin-rpc-redux-react`

- Refactor withRPCRedux HOC
- Remove legacy context (`ProviderPlugin` override)
- Migrate build to cupv4
- Rewrite tests (`tape-cup` => `jest`)

Legacy React context is no longer supported

I also deleted a significant amount from the README regarding how to register the plugin. Most of this was copied from the pass through plugin (`fusion-plugin-rpc`). I think the idea moving forward is that the `-react` packages we have will no longer be considered plugins since the idea of `ProviderPlugin` (a `fusion-react` plugin that wraps an existing Fusion.js plugin) is being removed. Instead, you will import tokens and plugins from the `fusion-plugin-` package and use them in React via some other Fusion.js package (e.g. `fusion-plugin-rpc-redux-react` - which we may want to rename in the future).


micburks:
Apologies for this getting huge. In adding tests, ended up switching to `cupv4`/`jest`. The logic for `node.test.js`, `browser.test.js`, and `interface.test.js` stayed the same, but they all run in node now. Looking into what tests might be duplicated or what else can be moved into the fixture I added.

fusion-plugin-rpc-redux-react/src/hook.js
```diff
@@ -0,0 +1,32 @@
...
+import {ReduxToken} from 'fusion-plugin-react-redux';
+import {createRPCHandler} from 'fusion-rpc-redux';
+
+type OptionalHandlerArgs = {
+  actions?: any,
+  transformParams?: (params: any) => any,
+  mapStateToParams?: (state: any, args?: any, ownProps: {}) => any,
+};
+
+export function useRPCHandler(rpcId: string, opts: OptionalHandlerArgs) {
```
ganemone: I think we should potentially hold off on exposing a useRPCHandler hook for a few reasons.

1. hooks cannot be used inside `prepared` functions. We could potentially refactor prepared to add support for this somehow (and maybe you already have that planned), but in the current state this doesn't work. This means that hooks could not be used for server side data fetching

2. I think the final API for this hook will look different when integrated into react suspense, which is the long term goal. I'm working on a proposal for what this could look like.

micburks: I'm interested to see what your ideas with Suspense are. I figured you might have a better idea of where to go with this idea. It was really just a helper, so I'll revert it.

I guess I thought hooks did work in async rendering. Did this PR (https://github.com/fusionjs/fusionjs/pull/227) not address this?

ganemone: That PR fixed hooks inside components during the prepare cycle, but not in the prepare function itself. I think the majority of this PR looks good, just want to hold off on the hook specifically.

micburks:
!merge

micburks:
!merge

PR #398, opened 2019-05-18
Use hooks in fusion-plugin-font-loader-react

angus-c:
Upgrade Context API via `useService`
Also use `useRef`, `useEffect` and `useState` in `withFontLoader`


fusion-plugin-font-loader-react/src/plugin.js
```diff
@@ -30,7 +27,15 @@ const plugin = createPlugin({
...
   },
+  provides: ({config}) => {
+    const {fonts, preloadDepth} = config;
+    const atomicFonts: AtomicFontsObjectType = (fonts: any);
+    const fallbackLookup = generateFallbackMap(atomicFonts, preloadDepth || 0);
+    const preloadSession = new PreloadSession(fallbackLookup);
+    return preloadSession.getFontDetails;
+  },
   middleware: ({config}) => {
+    // TODO: all this duplicate code
```
angus-c: whoops forgot to address this

PR #399, opened 2019-05-20
Use hooks in fusion-plugin-font-loader-react (take 2)

angus-c:


fusion-plugin-font-loader-react/package.json
```diff
@@ -26,6 +26,7 @@
     "./dist/browser.es2015.es.js": "./dist/browser.es2017.es.js"
   },
   "dependencies": {
+    "fusion-react": "0.0.0-monorepo",
```
micburks: I think you originally had this as a peerDep, any reason for moving it to deps?

fusion-plugin-font-loader-react/src/with-font-loading.js
```diff
@@ -27,57 +29,42 @@ All requested fonts should be defined in src/fonts/fontConfig.js
...
-            this.mounted &&
-              this.setState({$fontStyles: {fontFamily: fontName}});
+            mounted && setFontStyles({fontFamily: fontName});
           });
         }
-      }
+        return () => {
+          mounted.current = false;
+        };
+      });
```
micburks: I think without a second argument to `useEffect`, the effect will be run on every render of this component. This might be worth investigating and adding a `[fontName]` as the second parameter so it never recalculates since `fontName` never changes

fusion-plugin-font-loader-react/src/plugin.js
```diff
@@ -26,26 +23,38 @@ import type {
...
+    }
+    return preloadSession.getFontDetails;
+  },
   middleware: ({config}) => {
     const {fonts, preloadDepth, withStyleOverloads, preloadOverrides} = config;
     const atomicFonts: AtomicFontsObjectType = (fonts: any);
     const styledFonts: StyledFontsObjectType = (fonts: any);
-    const fallbackLookup = generateFallbackMap(atomicFonts, preloadDepth || 0);
-    const preloadSession = new PreloadSession(fallbackLookup);
+    if (!preloadSession) {
```
micburks: Can you get away with just pulling `getFontDetails` from the second parameter to `middleware`?

fusion-plugin-font-loader-react/package.json
angus-c: Yeah, I added it because the src code now has a direct dependence on `useService` in fusion-react

fusion-plugin-font-loader-react/src/with-font-loading.js
angus-c: Interesting, I'll validate and change if that's borne out

fusion-plugin-font-loader-react/src/plugin.js
angus-c: Sorry just seen this comment, not quite understanding. I don't think I use `getFontDetails` in `middleware`

micburks: Oh sorry, not `getFontDetails`, I meant `preloadSession`. I was just curious why you need to create the global `preloadSession` in both `provides` and `middleware` since it's the same logic. As far as I know, the `provides` logic will be run at startup, before we handle any requests with middleware.

angus-c: That's what I thought too, but  when I tested there seems to be a race condition. Should provides always be the second argument to middleware?

micburks: Thats weird. It seems like you should just be able to access `preloadSession` globally once you're in the middleware.

PR #400, opened 2019-05-20
Update create-fusion-app scaffold

micburks:
Update app scaffold `package.json`
- change name to app name when scaffolding
- add `private: true`
- remove `main` field, since apps has no entrypoint

### DIFF
--- a/create-fusion-app/bin/cli.js
+++ b/create-fusion-app/bin/cli.js
@@ -32,6 +32,7 @@ scaffold({
   cwd: path.join(__dirname, '..'),
   projectPath: path.join(process.cwd(), projectName),
   project: projectName,
+  packageJsonFields: {name: projectName},
 })
   .then(() => {
     console.log(`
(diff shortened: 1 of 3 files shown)


create-fusion-app/bin/cli.js
```diff
@@ -32,6 +32,7 @@ scaffold({
   cwd: path.join(__dirname, '..'),
   projectPath: path.join(process.cwd(), projectName),
   project: projectName,
+  packageJsonFields: {name: projectName},
```
AlexMSmithCA: Minor: are we moving away from leveraging Nunjucks for templating new Fusion.js projects?

micburks: I'm not aware of any plan. I think you're getting at why the `package.json` isn't just a Nunjucks file? Seems like it could be, but `fusion-scaffolder` already has this mechanism for replacing fields in the `package.json` so I just utilized that

AlexMSmithCA: Yeah, caught me a bit by surprise given we lean on Nunjucks for our internal templates.  🤷‍♂

LGTM!

micburks:
!merge

PR #401, opened 2019-05-20
Emit RPC stats from the browser

AlexMSmithCA:
Mirrored from https://github.com/fusionjs/fusion-plugin-rpc/pull/222:

> Currently stats are emitted on the server for each RPC call but my project now has the use case of understanding how long an RPC call lasts on the client (we want to measure the over the wire time).


AlexMSmithCA:
Closing as duplicate: https://github.com/fusionjs/fusionjs/pull/455

PR #402, opened 2019-05-20
Add hook for getting Redux store

micburks:
Add custom hook in `fusion-plugin-react-redux` for a simpler interface

The argument is that this is too much boilerplate for a super common task.
```js
import {FusionContext, useService} from 'fusion-react';
import {ReduxToken} from 'fusion-plugin-react-redux';
// ...
const ctx = React.useContext(FusionContext);
const reduxService = useService(ReduxToken);
const redux = reduxService.from(ctx);
const store = redux.store;
```

becomes

```js
import {useRedux} from 'fusion-plugin-react-redux';
// ...
const store = useRedux();
```

### DIFF
--- a/fusion-plugin-react-redux/package.json
+++ b/fusion-plugin-react-redux/package.json
@@ -42,6 +42,7 @@
     "execa": "^1.0.0",
     "flow-bin": "^0.98.1",
     "fusion-core": "0.0.0-monorepo",
+    "fusion-react": "0.0.0-monorepo",
     "fusion-test-utils": "0.0.0-monorepo",
     "nyc": "^14.1.0",
     "prettier": "^1.17.0",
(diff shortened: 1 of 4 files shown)


AlexMSmithCA:
Can we add documentation for this change?

micburks:
Closing this without merging for now. I'm not convinced this will be performant and I think I need to take it back to the drawing board

micburks:
I'm deciding I don't want to do this. I wanted this because `fusion-plugin-rpc-redux-react` needs access to the store to be able to access `store.dispatch`, but being a part of the `fusion-plugin-react-redux` API would suggest that this could be used in any consumer application to access state. I think the preferred method for subscribing to state will be the `connect` HOC due to vastly more functionality and performance optimizations.

PR #403, opened 2019-05-20
Add a tree shaking test that involves the use of Flow types

AlexMSmithCA:
Cloned from https://github.com/fusionjs/fusion-cli/pull/771:

> As part of the investigation into looking into WPT-2994 this PR adds a new test to ensure that imports/exports that include Flow types are properly stripped and tree shaken.


AlexMSmithCA:
Pushing a change that removes the offending code.  It appears unrelated to the Flow imports, and therefore unrelated to what we are testing here.

Seems like an edge case we may want to handle better in the future.

AlexMSmithCA:
!merge

PR #412, opened 2019-05-22
Fix dev command parameters

Monar:


Monar:
The issue here is to prevent minification for dev command. 
In here: 
```ts
minimizer: !dev && minify && runtime === 'client'
```
You just provide custom minimizer function and for false values it still uses the default one, the flag that controls minification is here:
```ts
 minimize: minify,
```
And because in compiler this [flag is set default to true](https://github.com/fusionjs/fusionjs/blob/master/fusion-cli/build/compiler.js#L137), and dev command didn't pass it then dev command is always running with some minification. 

So from what I understand to your match your expected behaviour, in stead of passing this minify flag I should change [this](https://github.com/fusionjs/fusionjs/blob/master/fusion-cli/build/get-webpack-config.js#L564) to this:

```ts
 minimize: !dev && minify,
```

Correct ?

KevinGrandon:
Hello, we looked into this and we believe that there was a bug introduced. Thank you for opening the PR, it helped us discover what's happening. Essentially we introduced minification to `dev` mode by default, which is definitely not what we want. Instead we plan on fixing this by disabling minification in dev mode. Please let us know if this doesn't address your issue, or if you have a use-case that isn't covered here.

The fix will be located in: https://github.com/fusionjs/fusionjs/pull/430

Monar:
@KevinGrandon yes this #430 is addressing my magnification needs but you also have bug in documentation vs code implementation: 
```
  - `--perserve-names`: Disable name mangling during script minification
```
I think you can remove this line from docs. :D

KevinGrandon:
> I think you can remove this line from docs. :D

Awesome, thank you again for pointing that out. Care to make a PR for this? :)

PR #416, opened 2019-05-23
Add script to auto update and commit lockfile

chrisdothtml:
This became more complex than I wanted, but the basic gist of it is that if `rush install` fails (due to out of date lockfile) in ci for a pull request, then it runs `rush update --full --purge` and commits the lockfile to the pull request branch.

Reasons for complexity:

- ci-gate causes lost info about the pull request from buildkite, so need to hit the github api to find branch for pull request (this added the most code with dependency-free utils, etc.)
- configuring local git with UberOpenSourceBot token, which is achievable via basic auth in the git remote url

Once all the git setup is done, the script itself is actually pretty straightforward - [common/scripts/rush-install-or-update.js](https://github.com/fusionjs/fusionjs/compare/ci-update-lockfile?expand=1#diff-b2778b976a2bf1c580c80ecd10ac9023)

### DIFF
--- a/common/scripts/configure-buildkite-git.js
+++ b/common/scripts/configure-buildkite-git.js
@@ -28,7 +28,7 @@ async function getRepoInfo() {
     branchName = BUILDKITE_BRANCH;
   } else {
     const [, pullRequestNumber] =
-      /^Pull request #(\d+)/.exec(BUILDKITE_MESSAGE) || [];
+      /^Pull request #(\d+)/i.exec(BUILDKITE_MESSAGE) || [];
     // this might fail if PR uses a fork and `maintainersCanModify` is false
     const pullRequest = await githubGraphql({
       token: GH_TOKEN,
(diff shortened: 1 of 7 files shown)


common/scripts/configure-buildkite-git.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+
+  if (!branchName || !fullRepoName) {
+    throw new Error(`Can't configure git; unable to get repo info`);
+  }
+
+  return {
+    branchName,
+    remoteUrl: `https://${GH_USERNAME}:${GH_TOKEN}@github.com/${fullRepoName}.git`,
```
chrisdothtml: alternative that doesn't need username ([ref](https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#http-based-git-access-by-an-installation)):

```sh
git clone https://x-access-token:<token>@github.com/owner/repo.git
```

PR #417, opened 2019-05-23
Refactor fusion-plugin-rpc-redux-react and fusion-plugin-connected-react-router to work with React-Redux 6 and up

zebralight:
* Address bug described [here](https://github.com/fusionjs/fusion-plugin-rpc-redux-react/issues/166)
* Do not include conditions to allow for React Redux <6


micburks:
Thanks for the PR!

We're currently working to remove React legacy context from all Fusion.js plugins. At least part of this work is a duplicate of an open PR (https://github.com/fusionjs/fusionjs/pull/397) that should fix the `react-redux` incompatibility in `fusion-plugin-rpc-redux-react` and permissively allow multiple versions of `react-redux`. I think that may be the best way to go for the time being until we decide if we want to skip `react-redux@6` and upgrade directly to version 7, rather than impart the performance regression introduced in version 6.

chrisdothtml:
Closing this. Our recommendation moving forward is to upgrade to react-redux v7. Feel free to open an issue if further discussion is needed on this 👍

PR #418, opened 2019-05-23
Add GetDataFromTreeToken

Monar:
With this token its possible to use [react-apollo-hooks](https://github.com/trojanowski/react-apollo-hooks#server-side-rendering) server side rendering. 

```
import {  GetDataFromTreeToken } from 'fusion-plugin-apollo';
import { renderToString } from 'react-dom/server';
import { getMarkupFromTree } from 'react-apollo-hooks';

const getDataFromTree = (tree) =>
  getMarkupFromTree({ renderFunction: renderToString, tree });

/* [...] */
  const app = new App(<Main />);
/* [...] */
  if(__NODE__) {
    app.register(GetDataFromTreeToken, getDataFromTree);
  }
  return app;
```


Monar:
@KevinGrandon When can we expected next release with this change ?

fusion-plugin-apollo/src/plugin.js
```diff
@@ -34,6 +35,7 @@ export type DepsType = {
   schema: typeof GraphQLSchemaToken.optional,
   endpoint: typeof GraphQLEndpointToken.optional,
   getApolloClient: typeof ApolloClientToken,
+  getDataFromTree: typeof GetDataFromTreeToken,
```
KevinGrandon: Looks like this just needs to be `typeof GetDataFromTreeToken.optional`. I'll go ahead and see if I can make that update inline.

Monar:
@KevinGrandon hey what are the perspectives of merging this PR?

KevinGrandon:
> @KevinGrandon hey what are the perspectives of merging this PR?

Sorry, we've had some infrastructure issues with our internal repos that we wanted to resolve with automation, but will do so manually since it shouldn't block our contributors. Let's land this, thanks a lot for the PR!

PR #420, opened 2019-05-23
Use scoped emitter when possible to emit browser stats

brendean:
Emit using a scope emitter (when possible) when the stats get re-emitted on the server (in our use case we want the request object to be able to determine the users location using a geo IP lookup).


KevinGrandon:
!merge

rtsao:
!merge

PR #421, opened 2019-05-24
Upgrade react-router-dom to v5.0

rtsao:
react-router 5 switches from the legacy context API to React.createContext, which breaks our usage of reading history from context in our Redirect component.

This PR is a bit of a stopgap, in the future, we should probably reconsider how router context is used and if/how/what we fork/extend/wrap from react-router. Ideally we can avoid using any private APIs and avoid similar problems in the future.


rtsao:
!merge

PR #422, opened 2019-05-24
Fix bug with default values for transform

ganemone:
The default value for transform should be 'all' if running on 'src/' code and 'spec' when running on node_modules (except for an exception in fusion-cli). This fixes a bug where the default was being set to 'spec' for all code.


ganemone:
!merge

Issue #327, opened 2019-05-09
[fusion-cli#427] Browser builds with node polyfills does not work with .mjs files


chrisdothtml:
This issue was migrated from [fusionjs/fusion-cli#427](https://github.com/fusionjs/fusion-cli/issues/427) and was originally reported by @ganemone.

---

Browser builds with node polyfills does not work with .mjs files

### Type of issue

Bug

### Description

If a browser file has an `.mjs` extension and uses a node builtin (e.g. process), and the project has the process polyfill configured, the compiled code will reference `require` rather than `__webpack_require__`. This causes an uncaught exception in the browser since `require` is not defined.

NOTE: There are likely many other issues with `.mjs` files. We should take verify that other aspects of the build process work with `.mjs` files.

### Current behavior

Generates code using require:

```js
/* WEBPACK VAR INJECTION */}.call(this, require("./../../node_modules/node-libs-browser/mock/process.js")))
```

### Expected behavior

Should generate code using __webpack_require__: 

```js
/* WEBPACK VAR INJECTION */}.call(this, __webpack_require__(/*! ./../node_modules/node-libs-browser/mock/process.js */ "./node_modules/node-libs-browser/mock/process.js")))
```

### Steps to reproduce

1. Scaffold a fusion project
2. Update the package.json to have: 
```json
{
  "node":  {
    "process": true,
    "Buffer": true
  }
}
```
3. Add a file: `test.mjs` with the following contents

```js
export default function test() {
  console.log(process)
}
```

4. import and use the test function in `src/main.js`

```js
import test from './test.mjs';
test();
```

5. run `fusion dev`

PR #423, opened 2019-05-24
Upgrade webpack to 4.28.4

rtsao:
Fixes https://github.com/fusionjs/fusion-cli/issues/427 https://github.com/fusionjs/fusionjs/issues/327

A couple notes:
- This fixes the erroneous `require`, but `process` still has to be imported (e.g. `import process from "process"`) 
- Webpack v5 has *completely* removed Node.js built-in polyfills aside from `__filename`, `__dirname`, and `global`. Accordingly, I don't think we should really consider this as part of the public API.
- Folks should probably not be using `.mjs` yet as module support is still experimental in Node.js

Webpack changes:
https://github.com/webpack/webpack/compare/v4.26.1...v4.28.4

Relevant change is https://github.com/webpack/webpack/pull/8511, released in `4.28.0`. This PR upgrades the the latest patch of this minor version.

### DIFF
--- a/common/config/rush/yarn.lock
+++ b/common/config/rush/yarn.lock
@@ -2,7 +2,7 @@
 # yarn lockfile v1
 
 
-"@apollographql/apollo-tools@^0.3.6", "@apollographql/apollo-tools@^0.3.6-alpha.1":
+"@apollographql/apollo-tools@^0.3.6":
   version "0.3.7"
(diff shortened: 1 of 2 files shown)


rtsao:
Superceded by https://github.com/fusionjs/fusionjs/commit/b599d8d8d78516a73ba81f221e84785747c59ad5

PR #430, opened 2019-05-28
Fix fusion-cli regression causing minification in development

ganemone:
Bug introduced in https://github.com/fusionjs/fusionjs/commit/36e3d9991a1b9fe5cecf0c6e5ca9982e769c48d2


rtsao:
!merge

PR #432, opened 2019-05-28
Fix IE11 bug in fusion-plugin-browser-performance-emitter

sumanvyj:
Fixes an error being thrown in IE11 by ensuring `toJSON` exists.


rtsao:
!merge

PR #439, opened 2019-05-30
Move ServiceContext Provider to root

micburks:
Adding ServiceContext.Provider from a plugin doesn't ensure that it is at the root of the tree. This context is special in that it should be accessible from other Providers that are added from plugin middleware


fusion-react/src/index.js
```diff
@@ -39,10 +38,24 @@ declare var __NODE__: Boolean;
...
   constructor(root: React.Element<*>, render: ?Render) {
-    if (!React.isValidElement(root))
+    if (!React.isValidElement(root)) {
       throw new Error(
         'Invalid React element. Ensure your root element is a React.Element (e.g. <Foo />) and not a React.Component (e.g. Foo)'
       );
+    }
+    const getService = token => {
+      // $FlowFixMe
+      const provides = this.getService(token);
```
micburks: Flow thinks we're accessing `this` before calling `super`

micburks:
!merge

PR #440, opened 2019-05-31
Add logic for limiting the number of events sent in each batch

ganemone:
koa-bodyparser puts a limit on the size of the request body it will parse to help prevent a DOS attack. If lots of events are sent at once from the universal-events plugin, the request body can exceed this limit. Since we have error handling logic that puts these events into local storage, this can put the user into an infinite cycle of failed requests as the events in local storage grow. 

This PR addresses this by putting a limit on the number of events which can be sent in a single batch. If the response fails with a 413 status code, we cut that limit in half for the current session.


ganemone:
!merge

PR #441, opened 2019-05-31
Move sourcemap install into entry files

ganemone:
This will allow for all imports to be affected by `experimentalBundleTest`, which is necessary for now.sh deployments to work.

### DIFF
--- a/fusion-cli/entries/server-entry.js
+++ b/fusion-cli/entries/server-entry.js
@@ -9,7 +9,8 @@
 /* eslint-env node */
 
 /* eslint-disable import/first */
-require('source-map-support').install();
+import sourceMapSupport from 'source-map-support';
+sourceMapSupport.install();
 
 // $FlowFixMe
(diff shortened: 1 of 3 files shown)


fusion-cli/entries/server-entry.js
```diff
@@ -8,6 +8,9 @@
 
 /* eslint-env node */
 
+/* eslint-disable import/first */
+require('source-map-support').install();
```
rtsao:
```suggestion
import sourceMapSupport from 'source-map-support';
sourceMapSupport.install();
```

fusion-cli/entries/serverless-entry.js
```diff
@@ -8,6 +8,9 @@
 
 /* eslint-env node */
 
+/* eslint-disable import/first */
+require('source-map-support').install();
```
rtsao:
```suggestion
import sourceMapSupport from 'source-map-support';
sourceMapSupport.install();
```

PR #443, opened 2019-06-03
RFC: Support asynchronous .provides

AlexMSmithCA:
[Rendered](https://github.com/fusionjs/fusionjs/blob/fa7f72cf85ba43d5db9a8e8972adfdd3b9700e0b/rfcs/text/0003-make-provides-async.md)

### DIFF
--- a/rfcs/text/0003-make-provides-async.md
+++ b/rfcs/text/0003-make-provides-async.md
@@ -153,7 +153,7 @@ TBD
 
 This approach involves a breaking change.  Plugin resolution now occurs asynchronously, which requires changes to `fusion-core` (breaking) and `fusion-cli` (non-breaking).
 
-Downstream effects include breaking changes to many consumers' test suites as well as testing helpers, like `fusion-test-utils`.  These often resolve the `FusionApp` themselves which must now be done asynchronously.
+Downstream effects include breaking changes to many consumers' test suites as well as testing helpers, like `fusion-test-utils`.  These often resolve the `FusionApp` themselves which must now be done asynchronously.  Many of these changes can be codemodded to avoid distruption to existing consumers, provided they follow best practices around testing for Fusion.js.
 
 # Alternatives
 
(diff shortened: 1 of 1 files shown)


rfcs/text/0003-make-provides-async.md
```diff
@@ -0,0 +1,160 @@
...
+    })
+  );
+
+  return app;
+};
+```
+
+# Motivation
+
+This proposal is motivated by situations where a plugin's service requires some initialization that occurs asynchronously (e.g. I/O, requests to back-end services, etc).  Today, there exists workarounds but they require additional boilerplate or are fairly clunky to use, leading to a degraded developer experience.
```
chrisdothtml: > there exists workarounds but they require additional boilerplate or are fairly clunky to use

Could you provide examples of this for those unfamiliar?

chrisdothtml: Nevermind, should have just kept reading lol

rfcs/text/0003-make-provides-async.md
```diff
@@ -0,0 +1,160 @@
...
+    return {
+      doSomething: () => {
+        /* some business logic */
+      }
+    };
+  }
+}));
+```
+
+A consuming plugin that depends upon the above plugin will be able to consume the exported service and use it synchronously.
```
chrisdothtml: So if I'm understanding correctly, this handles the `await` internally, so doesn't change the signature for using plugins, just makes it easier to author them?

AlexMSmithCA: Correct!  The signature for plugins would be unchanged.  It would, however, alter the behavior of an asynchronous service injected as a dependency.  In that case, the promise would be resolved for any consumers (versus today, where it would not).

chrisdothtml:
This looks good to me 👍. Probably could use a couple other eyes on it though

AlexMSmithCA:
!merge

PR #444, opened 2019-06-03
PoC: Support async .provides in fusion-core & fusion-cli

AlexMSmithCA:
Partial PoC for https://github.com/fusionjs/fusionjs/pull/443

### DIFF
--- a/fusion-cli/entries/client-entry.js
+++ b/fusion-cli/entries/client-entry.js
@@ -29,7 +29,7 @@ function reload() {
       );
       app.register(RoutePrefixToken, window.__ROUTE_PREFIX__);
     }
-    app.callback().call();
+    Promise.resolve(app.callback()).then(handler => handler.call());
   });
 }
 reload();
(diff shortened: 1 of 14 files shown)


AlexMSmithCA:
Closing this as it is not intended to land.  Proof-of-concept may be referenced or used in a future pull request.

brendean:
Is it likely Fusion will move forward with this RFC? My specific use case is initializing a GeoIP plugin service (using [maxmind](https://www.npmjs.com/package/maxmind)). Ideally we could wait for the DB to be loaded (async) only on start up of the app.

brendean:
I was able to unblock my use case using the [maxmind sync API](https://www.npmjs.com/package/maxmind#sync-api)

AlexMSmithCA:
As far as I know, we're still collecting use cases here and there is no concrete plan to support this right now.  I'll pass this back to the team.

As an alternative workaround - particularly if a sync API isn't available - is to do a one-time initialization in your plugin's middleware.  Not ideal as the first request will be slow, but future requests can reuse the initialized state.

PR #446, opened 2019-06-03
Add translation instrumentation to promises for split components

micburks:
- Annotate dynamic imports with the translation keys the chunk contains
- Update fusion-react to pass keys through splitComponentLoaders to i18n plugin
- Load translations in i18n plugins from keys instead of chunk ids

Fixes issues with dynamically loading translations where routing isn't guaranteed to hit the origin the initial request was served from. This will allow looking up a translation by key instead of the id of the chunk that contains it.

### DIFF
--- a/fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
+++ b/fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
@@ -30,6 +30,9 @@ type ClientPluginOpts = {
 };
 */
 
+const Dependency = require('webpack/lib/Dependency');
+const ImportPlugin = require('webpack/lib/dependencies/ImportPlugin');
+const ImportParserPlugin = require('webpack/lib/dependencies/ImportParserPlugin');
 const ImportDependency = require('webpack/lib/dependencies/ImportDependency');
 const ImportDependencyTemplate = require('webpack/lib/dependencies/ImportDependency')
   .Template;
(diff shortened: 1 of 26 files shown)


fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
```diff
@@ -56,71 +119,64 @@ class InstrumentedImportDependencyTemplate extends ImportDependencyTemplate {
...
+        new InstrumentedImportDependency.Template()
+      );
+      compilation.hooks.afterOptimizeDependencies.tap(name, modules => {
+        // Replace ImportDependency with our Instrumented dependency
+        for (const module of modules) {
+          if (module.blocks) {
+            module.blocks.forEach(block => {
+              block.dependencies.forEach((dep, index) => {
+                if (dep instanceof ImportDependency) {
+                  block.dependencies[index] = new InstrumentedImportDependency(
```
micburks: Replace `ImportDependency` with `InstrumentedImportDependency`

fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
```diff
@@ -9,13 +9,84 @@
...
+      chunkIds: this.chunkIds,
+      translationKeys: this.translationKeys,
+    };
+  }
+  updateHash(hash) {
+    super.updateHash(hash);
+    const {translationKeys} = this.getInstrumentation();
+    // Invalidate this dependency when the translation keys change
+    // Necessary for HMR
+    hash.update(JSON.stringify(translationKeys));
```
micburks: Dependencies only depend on module id by default. This logic will tell the dependency to update when `translationKeys` changes.

fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
```diff
@@ -48,80 +99,107 @@ class InstrumentedImportDependencyTemplate extends ImportDependencyTemplate {
...
     } else {
-      // client-side, use built-in values
-      chunkIds = getChunkGroupIds(depBlock.chunkGroup);
+      chunkIds = getChunkGroupIds(dep.block.chunkGroup);
     }
 
+    // This is a hack. Some production builds had undefined for the module.id
+    // which broke the build.
+    dep.module.id =
+      dep.module.id || (dep.module.identifier && dep.module.identifier());
```
micburks: I tried a lot of different hooks to find one where `module.id` would always be defined (such as `afterOptimizeModuleIds`, but no luck. Without this line, the production build worked in some fixtures but failed in others. I couldn't find any reason why it wasn't defined consistently.

fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
```diff
@@ -48,80 +99,107 @@ class InstrumentedImportDependencyTemplate extends ImportDependencyTemplate {
...
-    /**
-     * The standard plugin is added on `compile`,
-     * which sets the default value for `ImportDependency` in  the `dependencyTemplates` map.
-     * `make` is the subsequent lifeycle method, so we can override this value here.
-     */
-    compiler.hooks.make.tapAsync(name, (compilation, done) => {
-      if (this.clientChunkIndexState) {
-        // server
-        this.clientChunkIndexState.result.then(chunkIndex => {
+    if (this.opts.compilation === 'server') {
```
micburks: I ended up keeping the server compilation as it was before essentially. This promise makes it a little weird to work with, so I've kept the client and server compilations separate.

fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
```diff
@@ -48,80 +99,107 @@ class InstrumentedImportDependencyTemplate extends ImportDependencyTemplate {
...
-            new InstrumentedImportDependencyTemplate(chunkIndex)
+            new InstrumentedImportDependency.Template(metadata.fileManifest)
           );
           done();
         });
-      } else {
-        // client
+      });
+    }
+    if (this.opts.compilation === 'client') {
```
micburks: Only on the client build, add a new type of dependency. On the `afterOptimizeDependencies` hook, switch out all `Import` dependencies for `IntrumentedImport` dependencies

micburks:
!merge

PR #449, opened 2019-06-04
Remove orphaned-plugin exception

micburks:
- don't track nonPlugins
- remove error tests

With the implementation of `useService`, plugins that supply React context must be registered to a token. Previously, this wasn't required.

We currently have no way of tracking the relationship between plugins and components, it's purely enforced during rendering since `useService` will throw if the plugin was not registered. We couldn't suspend throwing an error until rendering is complete since not every component is rendered.

The only purpose this exception served was to make you aware of whether you needed a token or not. Supplying a token and not using it has no effect on the application's correctness.

Previous work:
- https://github.com/fusionjs/fusion-core/pull/331
- https://github.com/fusionjs/fusion-core/pull/119 (`old and busted, new hotness`)

### DIFF
--- a/fusion-core/src/__tests__/dependency-resolution.js
+++ b/fusion-core/src/__tests__/dependency-resolution.js
@@ -610,30 +610,6 @@ tape('error message when dependent plugin does not have token', t => {
   t.end();
 });
 
-tape('Extraneous dependencies', t => {
-  const app = new App('el', el => el);
-  const TestToken = createToken('test');
-  app.register(TestToken, 'some-value');
-  t.throws(() => app.resolve());
-  t.end();
-});
-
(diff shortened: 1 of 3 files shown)


rtsao:
@ganemone Do you remember a previous discussion about testing/mocking plugins where this enforcement was a consideration? Also I think there may have been one regarding plugin ordering too.

If we remove this, I wonder if we should re-visit those.

rtsao:
Bikeshed recap:
Accidentally registering server-only configuration on the client is a huge footgun this functionality is designed to prevent. Instead we should keep this as-is but recommend alternatives:

1. Wrap value in plugin instead of registering as plain value.
2. Import value directly (don't use Fusion DI)
3. Consume/Mock via React Context instead (don't use Fusion DI)

PR #450, opened 2019-06-04
[fusion-plugin-rpc-redux-react] Fix memory leak

phou87:
The current implementation has a memory leak. Each render, mapStateToParams gets wrapped in a new arrow function that calls the old mapStateToParams. The more insidious side effect of this is that this.context.store gets retained. On the server, this render function is called whenever new SSR markup is generated, and that markup happens to be retained by this.context.store (via the ctx field in the store). That means that every SSR string ever created since the app started will persist in memory!

Below are the heap snapshots from reloading a page powered by a Fusion app three times. You can see that the mapStateToParams stack grows larger each time. I clicked the "Collect Garbage" button before generating the snapshots, yet the three SSR strings remain.

bad_heap_1
bad_heap_2
bad_heap_3

Below is the heap snapshot after this fix. You can see that the blue spikes from earlier (representing the markup) are now cleaned up after clicking "Collect Garbage", and there is only one SSR string in memory now.

good_heap


ganemone:
!merge

PR #455, opened 2019-06-05
Add RPC stats from the browser

brendean:
Currently stats are emitted on the server for each RPC call but my project now has the use case of understanding how long an RPC call lasts on the client (we want to measure the over the wire time).


AlexMSmithCA:
Re-stamping from https://github.com/fusionjs/fusionjs/pull/401

fusion-plugin-rpc/src/types.js
```diff
@@ -19,7 +19,7 @@ import {
 type ExtractReturnType = <V>(() => V) => V;
 
 export type RPCDepsType = {
-  emitter?: typeof UniversalEventsToken,
+  emitter: typeof UniversalEventsToken,
```
AlexMSmithCA: Adding the `breaking` label for this.

AlexMSmithCA:
!merge

PR #456, opened 2019-06-05
Add config to fusion-plugin-rpc for customizing api path

bryandbor:
A config token is being added to the `fusion-plugin-rpc` plugin to allow for the customization of the path which the rpc endpoints call/listen. This means that rpc endpoints can be moved from the (default) path of `/api/...` to anything of the author's choosing. This config token is optional to retain full backward compatibility.


fusion-plugin-rpc/src/server.js
```diff
@@ -114,19 +119,26 @@ const pluginFactory: () => RPCPluginType = () =>
...
     middleware: deps => {
-      const {emitter, handlers, bodyParserOptions} = deps;
+      const {emitter, handlers, bodyParserOptions, rpcConfig} = deps;
       if (!handlers)
         throw new Error('Missing handlers registered to RPCHandlersToken');
       if (!emitter)
         throw new Error('Missing emitter registered to UniversalEventsToken');
       const parseBody = bodyparser(bodyParserOptions);
 
+      let apiPath = 'api';
```
micburks: Perhaps we should add some logic to handle whether `rpcConfig.apiPath` was supplied with leading/trailing `/`. Not sure how the server handles adjacent `/`, so could lead to a confusing issue to debug.

```suggestion
      let apiPath = rpcConfig && rpcConfig.apiPath
        ? rpcConfig.apiPath
        : 'api';
      apiPath = `/${apiPath}/`.replace(/\/+/, '/');
```

fusion-plugin-rpc/src/browser.js
```diff
@@ -41,11 +48,17 @@ class RPC {
       throw new Error('Missing emitter registered to UniversalEventsToken');
     }
     const fetch = this.fetch;
+    const config = this.config;
     const emitter = this.emitter;
+
+    let apiPath: string = 'api';
```
micburks: This logic could be moved to the constructor, rather than each request, especially if we add some logic to remove duplicate `/`

micburks:
!merge

PR #457, opened 2019-06-05
Upgrade legacy context [fusion-plugin-i18n-react]

micburks:
Upgrade legacy context

We still want to require users to register the plugin from this package so that the splitComponentLoaders callback is registered

### DIFF
--- a/fusion-plugin-i18n-react/src/with-translations.js
+++ b/fusion-plugin-i18n-react/src/with-translations.js
@@ -7,7 +7,6 @@
  */
 
 import React from 'react';
-import PropTypes from 'prop-types';
 import type {Context} from 'fusion-core';
 import {I18nContext} from './plugin.js';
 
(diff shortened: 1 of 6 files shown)


fusion-plugin-i18n-react/src/with-translations.js
```diff
@@ -61,9 +61,8 @@ export const withTranslations = (
 
     const displayName = Component.displayName || Component.name || 'Anonymous';
     WithTranslations.displayName = `withTranslations(${displayName})`;
-    WithTranslations.contextTypes = {
-      i18n: PropTypes.object.isRequired,
-    };
+    // $FlowFixMe
+    WithTranslations.contextType = I18nContext;
```
angus-c: curious what was the issue here

micburks: This was a surprise to me, too. I wasn't sure how to fix this
```
Cannot assign I18nContext to WithTranslations.contextType because property contextType is missing in statics of WithTranslations [1].
```

micburks:
!merge

PR #462, opened 2019-06-06
Support dynamic translations

### LINKED ISSUE #340: [fusion-cli#645] Improved Translation API via Hooks
This issue was migrated from [fusionjs/fusion-cli#645](https://github.com/fusionjs/fusion-cli/issues/645) and was originally reported by @ganemone.

---

We should investigate if we can improve our translations API using hooks.

### Type of issue

Feature Request

### Description

The current API for working with raw translated strings is a bit cumbersome, as it requires duplicating the translation keys, once in the `withTranslations` call, and once in the `translate` call. This is further described here: https://github.com/fusionjs/fusion-cli/pull/643

We should investigate various options of improving this API, potentially taking advantage of hooks.

### LINKED ISSUE #345: [fusion-cli#713] Support non-literal translation keys for i18n
This issue was migrated from [fusionjs/fusion-cli#713](https://github.com/fusionjs/fusion-cli/issues/713) and was originally reported by @schnerd.

---

### Type of issue

Feature request

### Problem

[fusion-plugin-i18n](https://fusionjs.com/api/fusion-plugin-i18n-react) currently only works with string literals:

```js
// Works
<Translate id="homepage.hero" />

// Doesn't work
const key = "homepage.hero";
<Translate id={key} />
```

This is because there is a [custom babel plugin](https://github.com/fusionjs/fusion-cli/blob/master/build/babel-plugins/babel-plugin-i18n/index.js) that extracts these strings at build time.

This limitation, however, leads to some very verbose and bloated logic in some cases. For example, when pulling translations for days of the week our old approach:

```js
return Locale.get(`time.weekdays.${day}`);
```

now turns into the following with Fusion

```js
  switch (dayNumber) {
    case 1:
      return <Translate id="time.weekdays.1" />;
    case 2:
      return <Translate id="time.weekdays.2" />;
    case 3:
      return <Translate id="time.weekdays.3" />;
    case 4:
      return <Translate id="time.weekdays.4" />;
    case 5:
      return <Translate id="time.weekdays.5" />;
    case 6:
      return <Translate id="time.weekdays.6" />;
    case 7:
      return <Translate id="time.weekdays.7" />;
    default:
      return '';
  }
```

The custom babel transform is awesome and seems necessary for efficient code splitting, but I'm hoping there's a strategy which gives us the best of both worlds.

### Potential Solution

One solution that comes to mind is a comment that acts as a directive to the i18n babel transform.

```js
<Translate id={
  // fusion-i18n: time.weekdays.*
  `time.weekdays.${id}`
} />
```

The transform would parse the comment and match it against keys found in the translation files. A asterisk wildcard match is probably sufficient, but in theory this could also be a regex if people wanted more flexibility.

This solution seems to offer the benefits of code splitting, without runtime overhead and code bloat.

Thoughts?

micburks:
Implemented based on this conversation https://github.com/fusionjs/fusionjs/issues/340

Closes #345
Closes #340

The `useTranslations` hook supports dynamic translations when given a hinted template string.

```js
const translate = useTranslations();

// 👍 works
translate('static')
translate(`static.${dynamic}`)
translate(`${dynamic}${dynamic}${dynamic}.static`)

// 👎 fails
translate(foo)
translate(`${foo}`)
translate(`${dynamic}${dynamic}${dynamic}`)
translate('static' + 'also-static')
translate(42)
```

The translation key extracted from `translate` is an array of strings. Using this array, we can match each item (in order) to the loaded translations. This prevents us from coming up with some dynamic regex or DSL pattern representation.

### DIFF
--- a/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
+++ b/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
@@ -57,7 +57,7 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
           const translationPaths =
             refPath.parentPath.scope.bindings[localName].referencePaths;
           translationPaths.forEach(translationPath => {
-            if (t.isCallExpression(translationPath.parentPath)) {
+            if (t.isCallExpression(translationPath.parentPath) && translationPath.parentKey === 'callee') {
               const arg = translationPath.parentPath.node.arguments[0];
               const errorMessage =
                 'useTranslations result function must be passed string literal or hinted template literal';
@@ -74,6 +74,8 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
               } else {
                 throw new Error(errorMessage);
               }
+            } else {
+              throw new Error('Unexpected usage of useTranslations return function');
             }
           });
         }
--- a/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
+++ b/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
@@ -51,13 +51,18 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
           });
         } else if (specifierName === 'useTranslations') {
           if (!t.isVariableDeclarator(refPath.parentPath.parent)) {
-            throw new Error('Unexpected assignment of useTranslations return function');
+            throw new Error(
+              'Unexpected assignment of useTranslations return function'
+            );
           }
           const localName = refPath.parentPath.parent.id.name;
           const translationPaths =
             refPath.parentPath.scope.bindings[localName].referencePaths;
           translationPaths.forEach(translationPath => {
-            if (t.isCallExpression(translationPath.parentPath) && translationPath.parentKey === 'callee') {
+            if (
+              t.isCallExpression(translationPath.parentPath) &&
+              translationPath.parentKey === 'callee'
+            ) {
               const arg = translationPath.parentPath.node.arguments[0];
               const errorMessage =
                 'useTranslations result function must be passed string literal or hinted template literal';
@@ -75,7 +80,9 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
                 throw new Error(errorMessage);
               }
             } else {
-              throw new Error('Unexpected usage of useTranslations return function');
+              throw new Error(
+                'Unexpected usage of useTranslations return function'
+              );
             }
           });
         }
--- a/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
+++ b/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
@@ -69,25 +69,12 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
               if (t.isStringLiteral(arg)) {
                 translationIds.add(arg.value);
               } else if (t.isTemplateLiteral(arg)) {
-                const result = arg.quasis
-                  .map(q => q.value.cooked)
-                  .filter((str, index) => {
-                    if (
-                      index !== 0 &&
-                      index !== arg.quasis.length - 1 &&
-                      str === ''
-                    ) {
-                      // interpolations in the middle of a string can be ignored
-                      // there is no guarantee that they will evaluate to a string of any length
-                      return false;
-                    }
-                    return true;
-                  });
-                if (result.join('') === '') {
+                const literalSections = arg.quasis.map(q => q.value.cooked);
+                if (literalSections.join('') === '') {
                   // template literal not hinted, i.e. translate(`${foo}`)
                   throw new Error(errorMessage);
                 } else {
-                  translationIds.add(result);
+                  translationIds.add(literalSections);
                 }
               } else {
                 throw new Error(errorMessage);
--- a/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
+++ b/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
@@ -69,7 +69,20 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
               if (t.isStringLiteral(arg)) {
                 translationIds.add(arg.value);
               } else if (t.isTemplateLiteral(arg)) {
-                const result = arg.quasis.map(q => q.value.cooked);
+                const result = arg.quasis
+                  .map(q => q.value.cooked)
+                  .filter((str, index) => {
+                    if (
+                      index !== 0 &&
+                      index !== arg.quasis.length - 1 &&
+                      str === ''
+                    ) {
+                      // interpolations in the middle of a string can be ignored
+                      // there is no guarantee that they will evaluate to a string of any length
+                      return false;
+                    }
+                    return true;
+                  });
                 if (result.join('') === '') {
                   // template literal not hinted, i.e. translate(`${foo}`)
                   throw new Error(errorMessage);
--- a/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
+++ b/fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
@@ -11,7 +11,7 @@
 const createModuleVisitor = require('../babel-plugin-utils/visit-named-module');
 
 const PACKAGE_NAME = ['fusion-plugin-i18n-react', 'fusion-plugin-i18n-preact'];
-const COMPONENT_IDENTIFIER = ['Translate', 'withTranslations'];
+const COMPONENT_IDENTIFIER = ['Translate', 'withTranslations', 'useTranslations'];
 
 module.exports = i18nPlugin;
 
@@ -43,6 +43,16 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
             if (!t.isStringLiteral(element)) {
               throw new Error(errorMessage);
             }
+            translationIds.add(element.value);
+          });
+        } else if (specifierName === 'useTranslations') {
+          const errorMessage =
+            'The useTranslations hook must be called with an array of strings';
+          if (!t.isArrayExpression(firstArg)) {
+            throw new Error(errorMessage);
+          }
+          const elements = firstArg.elements;
+          elements.forEach(element => {
             translationIds.add(element.value);
           });
         }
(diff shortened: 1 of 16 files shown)


fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
```diff
@@ -45,6 +49,30 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
             }
             translationIds.add(element.value);
           });
+        } else if (specifierName === 'useTranslations') {
+          const localName = refPath.parentPath.parent.id.name;
```
ganemone: Might want to validate the parentPath type is what you expect. For example, someone could do this 

```js
function MyComponent() {
  const obj = {};
  obj.translate = useTranslations();
}
```
or

```js
function MyComponent() {
  const obj = {};
  return <OtherComponent translate={useTranslations()} />
}
```

micburks: `useTranslations` can only be used as a CallExpression within a VariableDeclarator now

fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
```diff
@@ -57,7 +57,7 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
           const translationPaths =
             refPath.parentPath.scope.bindings[localName].referencePaths;
           translationPaths.forEach(translationPath => {
-            if (t.isCallExpression(translationPath.parentPath)) {
+            if (t.isCallExpression(translationPath.parentPath) && translationPath.parentKey === 'callee') {
```
micburks:
```js
translate('value')
// or
otherFunction(translate('value'))
```
but not
```js
otherFunction(translate)
```

fusion-plugin-i18n/src/node.js
```diff
@@ -22,6 +22,29 @@ import type {
   TranslationsObjectType,
 } from './types.js';
 
+function matchesOrder(key) {
+  return translation => {
+    let matchIndex = 0;
+
+    return key.every((part, i) => {
+      if (part === '') {
+        // part is result of two adjacent interpolations - skip
```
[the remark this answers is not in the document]
micburks: I believe it's more like this:
```js
translate(`abc${a}def`) // => ["abc", "def"]
translate(`abc${a}${b}def`) // => ["abc", "", "def"]
```

fusion-plugin-i18n/src/node.js
```diff
@@ -22,6 +22,29 @@ import type {
   TranslationsObjectType,
 } from './types.js';
 
+function matchesOrder(key) {
```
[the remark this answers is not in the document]
micburks: Yes will add more tests. There ended up being so many edge cases for other solutions that I decided to play it simple with the output of the static analysis and how this is matched against potential translations

fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
```diff
@@ -45,6 +49,42 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
...
+              t.isCallExpression(translationPath.parentPath) &&
+              translationPath.parentKey === 'callee'
+            ) {
+              const arg = translationPath.parentPath.node.arguments[0];
+              const errorMessage =
+                'useTranslations result function must be passed string literal or hinted template literal';
+              if (t.isStringLiteral(arg)) {
+                translationIds.add(arg.value);
+              } else if (t.isTemplateLiteral(arg)) {
+                const result = arg.quasis.map(q => q.value.raw);
```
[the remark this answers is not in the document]
rtsao: Oh, it seems this coalescing already happens in some cases, but not all?

```
`abc${foo}${bar}def${baz}`;
// 4 elements ["abc", "", "def", ""]

`${foo}${bar}`
// 3 elements
["", "", ""]
```

fusion-plugin-i18n/src/node.js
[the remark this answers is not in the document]
rtsao: I forgot how this works. The interpolations go in-between quasi elements.

fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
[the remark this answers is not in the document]
micburks: Yeah cooked is more accurate. Some examples:
```js
'a\bc' // => 'ac'
'a\\bc' // => 'a\bc'
'a\`bc' // => 'a`bc'
```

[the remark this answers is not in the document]
micburks: I think the reason I didn't want to completely ignore empty strings was because it is significant if they either start or end the match (meaning no `^` or `$` in regex terms). I think empty strings could be coalesced into a single empty string (or removed if in the middle of a match).

fusion-plugin-i18n/src/node.js
```diff
@@ -100,9 +134,21 @@ const pluginFactory: () => PluginType = () =>
           ctx.template.htmlAttrs['lang'] = localeCode;
         } else if (ctx.path === '/_translations') {
           const i18n = plugin.from(ctx);
-          const keys = querystring.parse(ctx.querystring).keys || '';
-          const translations = keys.split(',').reduce((acc, key) => {
-            acc[key] = i18n.translations && i18n.translations[key];
+          const keys = JSON.parse(
+            querystring.parse(ctx.querystring).keys || '[]'
```
micburks: Changed this to a stringified array to pass both strings and arrays to translations endpoint. No longer can break on `,`.

fusion-cli/build/babel-plugins/babel-plugin-i18n/test/index.test.js
```diff
@@ -53,3 +53,59 @@ export default function() {
     )
   ).toThrow('The translate component must have props.id be a string literal.');
 });
+
+test('babel-plugin-i18n - valid usage of useTranslations', () => {
```
micburks: Purpose of tests in this file:
- static analysis finds dynamic and static strings
- static analysis is added to Set
- error is thrown with improper usage of `translate`

fusion-cli/test/e2e/split-translations/test.js
```diff
@@ -29,6 +29,18 @@ test('`fusion build` app with split translations integration', async () => {
     content.includes('__MAIN_TRANSLATED__'),
     'app content contains translated main chunk'
   );
+  t.ok(
```
micburks: Purpose of tests in this file:
- i18n manifest (chunk-translation-map) properly includes dynamic translations arrays
- node plugin uses static analysis output to locate matching translations

fusion-plugin-i18n-react/src/__tests__/use-translations.test.browser.js
```diff
@@ -0,0 +1,30 @@
+// @flow
```
micburks: Purpose of tests in this file (no static analysis happens in these unit tests):
- `useTranslations` hook gets `translate` function from `I18nContext.Provider`
- `translate` is called properly

fusion-plugin-i18n/src/__tests__/index.browser.js
```diff
@@ -124,7 +124,7 @@ test('load', t => {
   };
   const data = {test: 'hello', interpolated: 'hi ${value}'};
   const fetch: any = (url, options) => {
-    t.equals(url, '/_translations?keys=test-key', 'url is ok');
+    t.equals(url, '/_translations?keys=["test-key"]', 'url is ok');
```
micburks: Purpose of tests in this file:
- uses proper over-the-wire format

fusion-plugin-i18n/src/__tests__/index.node.js
```diff
@@ -168,3 +168,68 @@ test('non matched route', async t => {
   t.notok(ctx.body, 'does not set ctx.body');
   t.end();
 });
+
+test('matchesOrder matches positionally', async t => {
```
micburks: Purpose of tests in this file:
- static analysis output is used accurately in a number of cases to match translations

fusion-cli/build/babel-plugins/babel-plugin-i18n/index.js
```diff
@@ -45,6 +49,55 @@ function i18nPlugin(babel /*: Object */, {translationIds} /*: PluginOpts */) {
...
+                const result = arg.quasis
+                  .map(q => q.value.cooked)
+                  .filter((str, index) => {
+                    if (
+                      index !== 0 &&
+                      index !== arg.quasis.length - 1 &&
+                      str === ''
+                    ) {
+                      // interpolations in the middle of a string can be ignored
+                      // there is no guarantee that they will evaluate to a string of any length
```
rtsao: I think I'm a bit confused by this comment.

Regarding the first part:
>  interpolations in the middle of a string can be ignored

My reading of this statement implies the substitution in (the middle of) `` `hello${interpolation}world` `` could be ignored. But in this case, the substitution seems rather crucial.

Regarding the second part:
> there is no guarantee that they will evaluate to a string of any length

Isn't this true of any interpolation? What makes interpolations in the middle of a string special?

Based on my interpretation of the code, this condition seems to be about filtering out redundant substitutions. This would happen in the case of adjacent substitutions. And unless I'm mistaken, I think that non-beginning and non-ending empty string literal sections only occur if and only if there's a pair of two adjacent substitutions.

micburks: > [...] non-beginning and non-ending empty string literal sections only occur if and only if there's a pair of two adjacent substitutions

You're correct, empty strings in the middle of the array would only occur with adjacent substitutions.

> Isn't this true of any interpolation? What makes interpolations in the middle of a string special?

This is true of any interpolation.

The substitution is significant, I meant to convey that the resulting empty quasis are insignificant if present in the middle of the array. Not just redundant but completely lacking of information. Filtering them out doesn't affect how we will match translations.

We have to treat these two statements as equivalent, since we don't know what will be in the substitution:
```js
`hello${''}${''}world`
`hello${' to the people '}${'of the '}world`
```
```json
{
  "helloworld": "matches",
  "hello to the people of the world": "matches"
}
```

That being said, regarding the other comments, I think removing the filtering would be the simplest approach.

fusion-plugin-i18n/src/__tests__/index.node.js
```diff
@@ -168,3 +168,68 @@ test('non matched route', async t => {
...
+    'cities.Chicago',
+    'cities.LosAngeles',
+    'animals.Buffalo',
+    'animals.Cat',
+    'test',
+  ];
+
+  // translate(`${}.Buffalo`);
+  // handles ending matches
+  const buffaloMatches = translations.filter(matchesOrder(['', '.Buffalo']));
```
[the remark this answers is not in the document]
micburks: Good idea 👍

micburks:
!merge

PR #463, opened 2019-06-06
Fix inconsistency between Flow and Prettier

AlexMSmithCA:
Resolves [WPT-3404](https://jeng.uberinternal.com/browse/WPT-3404):

> **Impetus**
> 
> It appears that the default lint rules for Fusion.js may cause an inconsistent state when typing multi-line Flow types.
> 
> **Reproduction steps**
> 
> An example is below:
> 
> ```js
> export type TestType = {
>   someProp:
>     | Array<number | string | {} | void | (() => void)>
>     | number
>     | string
>     | {}
>     | void
>     | (() => void),
> };
> ```
> 
> In the above case, ESlint will throw the following complaint:
> 
> > There must not be a line break after "someProp" type annotation colon.eslint(flowtype/space-after-type-colon)
> 
> The autogenerated fix for this results in:
> ```js
> export type TestType = {
>   someProp: | Array<number | string | {} | void | (() => void)>
>     | number
>     | string
>     | {}
>     | void
>     | (() => void),
> };
> ```
> 
> Which prettier complains about:
> 
> > Insert `⏎- - - `eslint(prettier/prettier)
> 
> These two ping pong back and forth.

### DIFF
--- a/eslint-config-fusion/index.js
+++ b/eslint-config-fusion/index.js
@@ -33,13 +33,7 @@ module.exports = {
 
     // Fix inconsistency between Flow (inherited rule from flowtype/recommended) and Prettier
     // https://jeng.uberinternal.com/browse/WPT-3404
-    'flowtype/space-after-type-colon': [
-      2,
-      'always',
-      {
-        allowLineBreak: true,
-      },
-    ],
+    'flowtype/space-after-type-colon': 'off',
   },
   settings: {
     react: {
--- a/eslint-config-fusion/index.js
+++ b/eslint-config-fusion/index.js
@@ -30,6 +30,16 @@ module.exports = {
     'react-hooks/rules-of-hooks': 'error',
     // https://github.com/facebook/react/issues/14920
     'react-hooks/exhaustive-deps': 'warn',
+
+    // Fix inconsistency between Flow (inherited rule from flowtype/recommended) and Prettier
+    // https://jeng.uberinternal.com/browse/WPT-3404
+    'flowtype/space-after-type-colon': [
+      2,
+      'always',
+      {
+        allowLineBreak: true,
+      },
+    ],
   },
   settings: {
     react: {


eslint-config-fusion/index.js
```diff
@@ -30,6 +30,16 @@ module.exports = {
     'react-hooks/rules-of-hooks': 'error',
     // https://github.com/facebook/react/issues/14920
     'react-hooks/exhaustive-deps': 'warn',
+
+    // Fix inconsistency between Flow (inherited rule from flowtype/recommended) and Prettier
```
AlexMSmithCA: Inherited rule:

https://github.com/gajus/eslint-plugin-flowtype/blob/712d840e0e72396a12a3e44021227f555306b75c/src/configs/recommended.json#L25-L28

The first option is whether to require a space or not after the type annotation colon (`"always"` in this case).

The second being whether a line break counts as a space following the annotation colon.  This one is key, as it's where the rule conflicts with Prettier.  If this is set to `"false"`, then when Prettier reformats the type to be on a new line, we conflict with this rule.

AlexMSmithCA:
Upon further inspection, there is still has some wonky behavior where we can still get stuck in a loop.
```js
type TestType = {
  prop:string, // no spaces
};
```
"fixes" to:
```js
type TestType = {
  prop:  string, // two spaces
};
```
which reverts back to no spaces.  Another loop.  In this case, however, the preferred format passes muster with both the linter and prettier, we just can't auto-fix the above issues.

Alternatively, we just turn off this rule and let Prettier handle the behavior around colons.  Thoughts?

KevinGrandon:
> Prettier handle the behavior around colons

This seems like a good option to me, does anything weird happen if we do this?

AlexMSmithCA:
Tested it out on `example-trips-viewer-fusion` with no issues.  FWIW, existing code should be fine, as it's already Prettier-ified.  Playing around with the types from above, they all work as expected.

🤷‍♂ Seems fine to me.

AlexMSmithCA:
!merge

Issue #247, opened 2019-04-30
[fusion-react#247] Portals support


rtsao:
Currently, client-side prepare is needed to support `prepared({defer: false})` (though worth noting that it isn't needed for `split()`.

To support hooks, a custom tree traversal was replaced with `react-dom/renderToString`.

However, this is problematic with Portals, because portals aren't supported in `react-dom/renderToString`.

**Short/medium term options**
- Userland workaround with conditional Portal rendering via `context.__IS_PREPARE__`
- Replace `react-dom/renderToString` in `prepare` with a custom, but hooks-supporting tree traversal such as https://github.com/FormidableLabs/react-ssr-prepass or https://github.com/Ephem/react-lightyear
- Change the API of `prepared` so that for `({defer: false})`, there must be a unique ID and/or it must be serializable/deserializable, allowing for removal of client-side `prepare` entirely. Or provide an easy way for serialization on the server and synchronous hydration on the client.

**Long-term options**
- Remove `prepared` and `split` entirely in favor of vanilla React Suspense APIs (e.g. `React.lazy`).

PR #471, opened 2019-06-13
Fix support for React Portals

micburks:
Fixes https://github.com/fusionjs/fusionjs/issues/247

Short-term fix for portal support in client-side prepare


common/config/rush/yarn.lock
```diff
@@ -1494,7 +1494,7 @@
 
 "@rush-temp/create-fusion-app@file:./projects/create-fusion-app.tgz":
   version "0.0.0"
-  resolved "file:./projects/create-fusion-app.tgz#3e4b87baaa08effcbc6e2c73442a5e2aa34b33cc"
```
mlmorg: What is this about?

micburks: After inspecting a little more, when I run `rush update` it removes the hashes, but when I run `rush update --full --purge` it adds the hashes.

@lhorie @rtsao What is expected here? Is this safe to ignore?

ganemone:
I'm onboard with this. I think the next step is we do a breaking change that removes support for client side prepare all together.

micburks:
!merge

PR #472, opened 2019-06-13
Add RFC for enhanced Flow support in the monorepo

AlexMSmithCA:
[Rendered](https://github.com/fusionjs/fusionjs/blob/d06bb4e244a541ad8e061098a24f68b2e7917001/rfcs/text/0004-support-flow-in-monorepo.md)


rfcs/text/0004-support-flow-in-monorepo.md
```diff
@@ -0,0 +1,58 @@
...
+
+Furthermore, Flow resolves libdef files in the `<PROJECT_ROOT>/flow-typed/` directory by default.  Given that each project has its own `flow-typed` directory, and no project is at the "root" anymore, Flow fails to resolve any of these when running the `rush flow` command.  Unfortunately, Flow will not complain about missing type information, instead opting to resolve all untyped imports to `any`.  This may cause type coverage regressions.
+
+# Solution
+
+Align on a single `.flowconfig` file to live in the root of the monorepo.  While one exists today, it importantly misses many of the library definition (libdef) files defined by individual projects' configurations.
+
+We must ensure all relevant libdef files are included in the root configuration.  Any custom configuration in a single project must also be either supported across the entire monorepo, or removed.
+
+If multiple projects require a libdef for the same dependency, we ideally house only a single source of truth for that type information.  In this way, the monorepo solution cuts down on duplicated code and ensures consistency across projects.
```
micburks: I thought part of the problem was also that we publish the libdefs along with any package that uses them. How would we handle publishing if the flow-typed directory exists outside the package?

AlexMSmithCA: Good point!  In theory, we probably should not be publishing packages with libdef files.  Best practice would be that our packages are either entirely self-contained (i.e. no need for any definitions from a libdef file) or rely only upon open-sourced definitions that can be installed directly from [flow-typed](https://github.com/flow-typed/flow-typed).

Happy to bikeshed this as well!

rfcs/text/0004-support-flow-in-monorepo.md
```diff
@@ -0,0 +1,58 @@
...
+.*/test-artifacts/.*
+```
+
+Whereas [`fusion-tokens`](https://github.com/fusionjs/fusionjs/blob/master/fusion-tokens/.flowconfig) defines none.
+
+Furthermore, Flow resolves libdef files in the `<PROJECT_ROOT>/flow-typed/` directory by default.  Given that each project has its own `flow-typed` directory, and no project is at the "root" anymore, Flow fails to resolve any of these when running the `rush flow` command.  Unfortunately, Flow will not complain about missing type information, instead opting to resolve all untyped imports to `any`.  This may cause type coverage regressions.
+
+# Solution
+
+Align on a single `.flowconfig` file to live in the root of the monorepo.  While one exists today, it importantly misses many of the library definition (libdef) files defined by individual projects' configurations.
```
ganemone: I think this works well for our use case, but I imagine this could be difficult to apply to larger monorepos (like a monorepo of services). The maps/web-monorepo today has a single flow config and starting up the flow server is very slow.

rfcs/text/0004-support-flow-in-monorepo.md
KevinGrandon: If we do get rid of this, we should look for an improvement in the process, like validating or automatically installing the correct libdefs.

AlexMSmithCA: Yup, I think we'd probably want explore to install libdefs as part of scaffolding.

rfcs/text/0004-support-flow-in-monorepo.md
AlexMSmithCA: Agreed, this is something we should keep in mind.

Flow and VSCode actually work fairly well in the scenario where we have mirrored `.flowconfig`s per project, described below.  In that case, Flow lazily starts a new server upon opening a Flow-enabled file in a project with its own `.flowconfig`.  Trade-off being potentially more Flow servers running, but they're more limited in scope.

rfcs/text/0004-support-flow-in-monorepo.md
```diff
@@ -0,0 +1,58 @@
...
+
+We must ensure all relevant libdef files are included in the root configuration.  Any custom configuration in a single project must also be either supported across the entire monorepo, or removed.
+
+If multiple projects require a libdef for the same dependency, we ideally house only a single source of truth for that type information.  In this way, the monorepo solution cuts down on duplicated code and ensures consistency across projects.
+
+Individual projects will have the same `.flowconfig` file that points to monorepo libdef files.  This allows folks to continue to run `yarn flow` within a single project during development.
```
albertywu: So does the proposed workflow look something like this:

1. user opens individual project
2. user realizes they need to pull in 3rd party typings for `foo@1.2.3`
3. `(cd .. && flow-typed install foo@1.2.3)`

?

ganemone: That sounds correct.

AlexMSmithCA: Yes, we should be installing new libdefs in the root of the monorepo(s).  If a libdef is needed in *both* the public and private monorepo, installing it in the root of the public one allows access in both.

PR #477, opened 2019-06-15
Fix client url in context [fusion-core]

micburks:
Must have been a typo. This doesn't seem to be used anywhere, but was the string `"undefined"` since a `undefined` and a string were being concatenated.


micburks:
!merge

PR #480, opened 2019-06-18
Replace vendored tree shake babel plugin

rtsao:
The [standalone package](https://www.npmjs.com/package/babel-plugin-transform-prune-unused-imports) includes some bugfixes and has a much better test suite


rtsao:
!merge

PR #490, opened 2019-06-20
Fix split component in simulation tests

rtsao:
We cannot assume `__webpack_modules__` is defined (i.e. Jest simulation tests that aren't bundled via Webpack)


rtsao:
!merge

PR #495, opened 2019-06-25
Implement rate limiting for fusion-plugin-error-handler

spyro92:
We are seeing spikes in fusion client errors due to single users spamming the _errors endpoint. So I limited the amount of errors sent to the backend of each type to 3.


spyro92:
@chrisdothtml Our client error dashboards have huge spikes where single users spam the system with hundreds of occurrences of the exact same error

We are addressing this problem the same way @ganemone did on muber

Happy to put it behind a config option, but I think its better as a default

ganemone:
I think this is probably fine as a default. It only restricts errors from a single user which have the exact same message.

chrisdothtml:
@spyro92 published in `0.0.0-canary.407b8e4.0`

PR #507, opened 2019-06-28
Audit libdefs and update to @latest

AlexMSmithCA:
Audit involves:
* Ensuring that relevant libdefs at up-to-date
* If custom edits have been made, move these out of the `flow-typed/npm/` directory which should be reserved for libdefs directly from [flow-typed](https://github.com/flow-typed/flow-typed).

In the future, we should look to recontribute our changes back to flow-typed.

Overall coverage has increased marginally:

image

(from [6315 lines uncovered](https://user-images.githubusercontent.com/3497835/60376676-6b574800-99c5-11e9-8c07-e339c94ed4eb.png))


fusion-plugin-universal-logger/src/tokens.js
```diff
@@ -6,11 +6,11 @@
  * @flow
  */
 
-import type {$winstonLoggerConfig} from 'winston';
+import type {LoggerConfig} from 'winston';
```
AlexMSmithCA: These types weren't previously exported by the `winston` libdef, and we had made custom edits to export the raw private types.

Consumers that use our altered libdefs will want to make this same change.

PR #509, opened 2019-07-02
Declare transitive dependencies to prevent warnings

micburks:
Yarn will warn about some implicit transitive dependencies. This can be fixed by making them actual dependencies of Fusion

This PR has private changes as well

### DIFF
--- a/eslint-config-fusion/package.json
+++ b/eslint-config-fusion/package.json
@@ -19,9 +19,9 @@
   "peerDependencies": {
     "eslint-plugin-flowtype": "^3.8.1",
     "eslint-plugin-import": "^2.17.2",
-    "eslint-plugin-jest": "^22.5.1",
+    "eslint-plugin-jest": "^22.7.1",
     "eslint-plugin-react": "^7.13.0",
-    "eslint-plugin-react-hooks": "^1.6.0"
+    "eslint-plugin-react-hooks": "^1.6.1"
   },
   "devDependencies": {
     "babel-eslint": "^10.0.1",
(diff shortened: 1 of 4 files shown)


fusion-cli/package.json
```diff
@@ -101,7 +101,13 @@
   },
   "peerDependencies": {
     "fusion-core": "0.0.0-monorepo",
-    "fusion-tokens": "0.0.0-monorepo"
+    "fusion-tokens": "0.0.0-monorepo",
+    "graphql": "^14.1.1"
+  },
+  "peerDependenciesMeta": {
```
micburks: This is a feature added in 1.13.0 of yarn and is undocumented as far as I can tell. The effect is that no warning will be emitted when peerDep graphql is not found

eslint-config-fusion/package.json
```diff
@@ -19,20 +19,20 @@
   "peerDependencies": {
```
micburks: This file is just dependency updates

fusion-react/package.json
```diff
@@ -40,7 +40,8 @@
   },
   "dependencies": {
     "@rtsao/react-ssr-prepass": "^1.0.5",
-    "prop-types": "^15.7.2"
+    "prop-types": "^15.7.2",
+    "react-is": "^16.8.0"
```
micburks: `react-is` is a peerDependency of `react-ssr-prepass`. By adding it here, it is no longer required for the consumer to install it

micburks:
!resync

micburks:
!resync

PR #512, opened 2019-07-02
Test that i18n has correct properties

micburks:
Default value for I18nContext is and empty object. Need to refine check here


micburks:
!merge

PR #515, opened 2019-07-04
Fix peer dep warnings

micburks:
Yarn will warn about some implicit transitive dependencies. This can be fixed by making them actual dependencies of Fusion

This PR has private changes as well

Reopening from https://github.com/fusionjs/fusionjs/pull/509


micburks:
!merge

PR #518, opened 2019-07-08
Move Babel loader into Jest worker farm

UberOpenSourceBot:
Move the babel loader into a jest worker farm so it can make use of more cores. Also, some small refactoring.


rtsao:
!resync

PR #532, opened 2019-07-12
Fix types for fusion-plugin-react-router

AlexMSmithCA:
Changes include:
* Updates to `types.js` to mirror [flow-typed/.../react-router-dom_v4.x.x.js](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-router-dom_v4.x.x/flow_v0.98.x-/react-router-dom_v4.x.x.js)
* Minor de-dup of existing types

### DIFF
--- a/fusion-plugin-react-router/src/modules/BrowserRouter.js
+++ b/fusion-plugin-react-router/src/modules/BrowserRouter.js
@@ -8,23 +8,13 @@
 
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import {Router as BaseRouterUntyped} from 'react-router-dom';
+import {Router as BaseRouter} from 'react-router-dom';
 
-import type {RouterType, RouterHistoryType} from '../types.js';
+import type {RouterPropsType as PropsType, RouterType} from '../types.js';
 
 export {Status, NotFound} from './Status.js';
 export {Redirect} from './Redirect.js';
 
-const BaseRouter: RouterType = (BaseRouterUntyped: any);
-
-type PropsType = {|
-  context?: any,
-  onRoute?: Function,
-  history: RouterHistoryType,
-  Provider?: RouterType,
-  basename?: string,
-  children?: React.Node,
-|};
 type ContextType = {
   __IS_PREPARE__: boolean,
 };
@@ -58,6 +48,7 @@ class BrowserRouter extends React.Component<PropsType> {
     const {Provider, history, basename} = this.props;
     if (!Provider) throw new Error('Missing Provider for Browser Router');
     return (
+      // $FlowFixMe
       <Provider basename={basename} history={history}>
         {this.props.children}
       </Provider>
@@ -81,5 +72,5 @@ BrowserRouter.childContextTypes = {
   onRoute: PropTypes.func.isRequired,
 };
 
-const BrowserRouterTyped: React.ComponentType<PropsType> = BrowserRouter;
+const BrowserRouterTyped: RouterType = BrowserRouter;
 export {BrowserRouterTyped as Router};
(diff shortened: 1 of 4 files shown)


fusion-plugin-react-router/src/modules/BrowserRouter.js
```diff
@@ -58,6 +48,7 @@ class BrowserRouter extends React.Component<PropsType> {
     const {Provider, history, basename} = this.props;
     if (!Provider) throw new Error('Missing Provider for Browser Router');
     return (
+      // $FlowFixMe
```
AlexMSmithCA: Necessary to resolve the following Flow complaint:

```sh
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/modules/BrowserRouter.js:51:8

Cannot create Provider element because property basename is missing in object type [1] but exists in props [2].

     src/modules/BrowserRouter.js
      48│     const {Provider, history, basename} = this.props;
      49│     if (!Provider) throw new Error('Missing Provider for Browser Router');
      50│     return (
 [2]  51│       <Provider basename={basename} history={history}>
      52│         {this.props.children}
      53│       </Provider>
      54│     );
      55│   }
      56│ }

     src/types.js
 [1] 148│ type BaseRouterType = React.ComponentType<{|
     149│   history: RouterHistoryType,
     150│   children?: React.Node,
     151│ |}>;
```

Note that`BaseRouterType` is taken directly from the libdef on flow-typed:

https://github.com/flow-typed/flow-typed/blob/c65b1c9cdf830e095ebc359130f1d3330428cf51/definitions/npm/react-router-dom_v4.x.x/flow_v0.98.x-/react-router-dom_v4.x.x.js#L121-L124

It isn't clear to me if `basename` should be added.  Happy to hear others' thoughts here.

fusion-plugin-react-router/src/modules/ServerRouter.js
```diff
@@ -51,6 +40,7 @@ class ServerRouter extends React.Component<PropsType> {
     const {Provider, history, basename, children} = this.props;
     if (!Provider) throw new Error('Missing Provider for Server Router');
     return (
+      // $FlowFixMe
```
AlexMSmithCA: Same justification as https://github.com/fusionjs/fusionjs/pull/532/files#r303153458

fusion-plugin-react-router/src/types.js
```diff
@@ -8,6 +8,24 @@
...
+  onRoute?: Function,
+  history: RouterHistoryType,
+  Provider?: BaseRouterType,
+  basename?: string,
+  children?: React.Node,
+};
+
+export type RouterType = React.ComponentType<RouterPropsType>;
+
+/* Types below adapted from flow-typed's libdef for react-router-dom
```
AlexMSmithCA: Updated all of the below types to mirror those from the libdef.  Ideally, these remain in sync.

fusion-plugin-react-router/src/modules/BrowserRouter.js
micburks: This is definitely confusing. It looks like the base `Router` takes history and children props (https://reacttraining.com/react-router/web/api/Router), but if the user was to supply their own router: https://github.com/fusionjs/fusionjs/blob/d25d2c6ada528d153ae1bda41a16be2c9bbce885/fusion-plugin-react-router/src/plugin.js#L55 then basename should be supplied because it could be one of the other router types (`BrowserRouter`, `HashRouter`, etc.) that do take `basename` prop

AlexMSmithCA:
!resync

AlexMSmithCA:
!merge

PR #537, opened 2019-07-14
Use custom module ids for instrumented module imports

micburks:
There are 2 issues that are somewhat related, instrumented module ids and instrumented chunk ids. I grouped these changes together because the issues can be reproduced in the same fashion (dynamically importing code that imports other code).

### Issue - Module IDs
Some production builds were getting these errors:

`Bundle does not contain a default export ` from [here](https://github.com/fusionjs/fusionjs/blob/master/fusion-react/src/async/split.js#L108)

or

`Cannot read property 'call' of undefined` because `modules[moduleId].call(/* ... */)` in webpack bootstrap code.

### Cause
This seems to have been because the instrumented `__MODULE_ID` was different than the `__webpack_modules__` key.

### Fix
The initial implementation of the translation instrumentation, I was unsure why some production builds were missing a `module.id`, but it appears to be because `afterOptimizeDependencies` is an earlier compilation phase then `beforeModuleIds ` where the module ids are assigned in HashedModuleIdsPlugin, which is only used in production builds for client bundles.

This fix creates custom module ids at the `afterOptimizeDependencies` phase, and makes sure they are reused in the `beforeModuleIds` phase by caching them based on module path (from `module.libIdent`).

### Issue - Chunk IDs
In some places, translations were not working correctly. Serialized translations were empty or incomplete and `/_translations` API endpoint was responding with an empty object.

### Cause
This is caused by a disconnect between chunk ids on the server versus the client for the same module.

### Fix
Use `module.rootModule.request` rather than `module.originRequest` for `ConcatenatedModule`s in the production build.


micburks:
> How difficult would it be to create an e2e regression test case for the bug this fixes?

I'll try again. Up until I found this fix, I was unable to reproduce the bug in simple apps

fusion-cli/test/e2e/split-translations/fixture/src/split1.js
```diff
@@ -2,9 +2,11 @@
 
 import React from 'react';
 import {Translate} from 'fusion-plugin-i18n-react';
+import OtherComponent from './other-component.js';
```
micburks: This seems a little crazy that importing another component would break the build, but it does. It's possible this becomes a different type of `AsyncBlock` dependency like a `ConcatenatedModule` or something when importing code that doesn't exist in a different bundle.

On master:
The module id in the instrumentation is an absolute path to a file, but the module id is the hash `fPq8`.

On this branch:
The module id in the instrumentation and the chunk are both `__fusion__0`.

micburks:
> > How difficult would it be to create an e2e regression test case for the bug this fixes?
> 
> I'll try again. Up until I found this fix, I was unable to reproduce the bug in simple apps

Nvm. This was super easy to reproduce

fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
```diff
@@ -108,19 +115,19 @@ InstrumentedImportDependency.Template = class InstrumentedImportDependencyTempla
     } else if (this.clientChunkIndex) {
       // Template invoked without InstrumentedImportDependency
       // server-side, use values from client bundle
-      let ids = this.clientChunkIndex.get(
-        (dep.module && dep.module.resource) || dep.originModule.resource
+      const ids = this.clientChunkIndex.get(
+        dep.module instanceof ConcatenatedModule
```
micburks: With the translations instrumentation, the chunk ids got out of sync between server and client build. This caused the i18n node plugin to lookup chunk ids that apparently had no translations.

The fix here is to use `dep.module.rootModule` for `ConcatenatedModules` instead of `dep.originModule` because the latter refers to the *importing* module, whereas the former refers to the *imported* module.

fusion-cli/test/e2e/split/test.js
```diff
@@ -16,10 +17,21 @@ test('`fusion dev` CHUNK_ID instrumentation', async () => {
...
-  const res = await request(`${url}/test`);
-  t.deepEqual(JSON.parse(res), [0]);
-  t.deepEqual(JSON.parse(resA), [0, 1]);
-  t.deepEqual(JSON.parse(resB), [0, 2]);
+  const resCombined = await request(`${url}/test-combined`);
+  const resTransitive = await request(`${url}/test-transitive`);
+  t.deepEqual(JSON.parse(resA), [0, 2]);
+  t.deepEqual(JSON.parse(resB), [0, 3]);
+  t.deepEqual(JSON.parse(resCombined), [0]);
+  t.deepEqual(JSON.parse(resTransitive), [1]);
```
micburks: I added another top-level dependency (`"test-d"`), so these chunk ids got bumped:
chunk 0 contains `"test-a"`, `"test-b"`, and `"test-c"`
chunk 1 contains `"test-d"`
chunk 2 contains `"test-a"`
chunk 3 contains `"test-b"`

fusion-cli/test/e2e/split/test.js
```diff
@@ -33,9 +45,20 @@ test('`fusion build` with dynamic imports and group chunks', async () => {
...
   const resA = await request(`http://localhost:${port}/test-a`);
   const resB = await request(`http://localhost:${port}/test-b`);
-  const res = await request(`http://localhost:${port}/test`);
-  t.deepEqual(JSON.parse(res), [10003, 3]);
+  const resCombined = await request(`http://localhost:${port}/test-combined`);
+  const resTransitive = await request(`http://localhost:${port}/test-transitive`);
   t.deepEqual(JSON.parse(resA), [10003, 10004, 3, 4]);
   t.deepEqual(JSON.parse(resB), [10003, 10005, 3, 5]);
+  t.deepEqual(JSON.parse(resCombined), [10003, 3]);
+  t.deepEqual(JSON.parse(resTransitive), [10006, 6]);
```
micburks: This fails on master because `"test-d"` erroneously thinks it shows up in chunks 0, 1, and 2 (10000, 10001, and 10002 in legacy build)

micburks:
!resync

micburks:
!resync

PR #540, opened 2019-07-15
Catch parseBody errors

brendean:
Currently if the call to `parseBody` fails (either because the body is too big for malformed JSON) the exception is not caught. I think we should catch the exception here and then return a failure response without trying to execute the handler.


AlexMSmithCA:
!merge

PR #548, opened 2019-07-16
Export all types from fusion-plugin-react-redux

AlexMSmithCA:
Resolves [WPT-3794](https://jeng.uberinternal.com/browse/WPT-3794):

> We'd like as much typing as possible.


AlexMSmithCA:
!merge

PR #550, opened 2019-07-17
Clean up typing for Tokens

phou87:
The old way that we extracted the underlying generic service type (T) that a token (Token<T>) provides was as follows:

- We defined the Token type to be a callable function that returns T (even though it isn't one!)
- We used an any cast to coerce created tokens into this type
- We used ExtractReturnType along with $Call to "call" the token type itself

This is fairly unintuitive and roundabout, when we can simply fetch the generic type directly by adjusting the utility function. I also exported it so that others can use it, since not all plugins export their underlying types.

### DIFF
--- a/fusion-core/src/base-app.js.flow
+++ b/fusion-core/src/base-app.js.flow
@@ -9,7 +9,7 @@
 import type {
   aliaser,
   cleanupFn,
-  ExtractReturnType,
+  ExtractTokenType,
   FusionPlugin,
   Middleware,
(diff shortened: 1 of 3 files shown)


KevinGrandon:
!merge

AlexMSmithCA:
It looks like this `!merge` did not trigger the sync'd PR to land:  https://github.com/uber/fusionjs/pull/301

Going to manually update and land it.

Issue #261, opened 2019-05-09
[fusion-plugin-service-worker#25] Handle SW middleware errors


chrisdothtml:
This issue was migrated from [fusionjs/fusion-plugin-service-worker#25](https://github.com/fusionjs/fusion-plugin-service-worker/issues/25) and was originally reported by @angus-c.

---

PR #553, opened 2019-07-19
Send 500 on Service Worker middleware error

angus-c:
Resolves fusionjs/fusionjs#261 

Previously we just logged the error.
I spent too long trying to write a test for this but it's actually really hard to get it to fail without editing non-test code.

### DIFF
--- a/fusion-plugin-service-worker/src/server.js
+++ b/fusion-plugin-service-worker/src/server.js
@@ -61,7 +61,7 @@ export default ((__NODE__ &&
             } catch (err) {
               ctx.status = 500;
               ctx.body = {error: err.message};
-              // logger.error('Error in Service Worker endpoint:', err);
+              logger.error('Error in Service Worker endpoint:', err);
             }
           }
           return next();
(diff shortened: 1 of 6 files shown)


fusion-plugin-service-worker/src/server.js
```diff
@@ -54,8 +54,11 @@ export default ((__NODE__ &&
                   : [],
                 cacheDuration,
               });
-            } catch (e) {
-              console.log('Error in Service Worker endpoint:', e); // eslint-disable-line
+            } catch (err) {
+              ctx.status = 500;
+              ctx.body = {error: err.message};
+              // log here because `sw.register` will not report the err object back to the client
+              console.log('Error in Service Worker endpoint:', err); // eslint-disable-line
```
mlmorg: Should this be depending on and using the LoggerToken instead of console.log?

angus-c: Fixed

angus-c:
reverted

PR #554, opened 2019-07-19
Send 500 and log error

angus-c:


fusion-plugin-service-worker/fixture-apps/app/src/main.js
```diff
@@ -16,6 +16,16 @@ import MockErrorPlugin from './plugins/mock-server-error';
 
 import root from './root.js';
 
+const createMockLogger = () => ({
```
AlexMSmithCA: We should probably just export a mock logger from `fusion-tokens`.  We have quite a few different mock implementations floating around now.

fusion-plugin-service-worker/src/__tests__/utils.node.js
```diff
@@ -101,3 +101,13 @@ export async function logCacheDates(page: Page, label: string) {
     );
   }, label);
 }
+
+export const createMockLogger = () => ({
```
AlexMSmithCA: Duplicated code?

fusion-plugin-service-worker/fixture-apps/app/src/main.js
angus-c: yeah, fixture app failed jmporting types import from `fusion-plugin-service-worker/src/__tests__/utils.node.js`

PR #560, opened 2019-07-22
Handle unparseable UA string

angus-c:


fusion-cli/plugins/ssr-plugin.js
```diff
@@ -173,5 +173,10 @@ function checkModuleSupport({name, version}) {
 }
 
 function majorVersion(version) {
-  return parseInt(version.split('.')[0], 10);
+  // version can be undefined, see https://jeng.uberinternal.com/browse/WPT-3893
+  if (typeof version === 'string') {
+    return parseInt(version.split('.')[0], 10);
+  } else {
+    return 0; // not sure if this is optimal but better than erroring
```
rtsao: IMO coercing undefined to 0 is counterintuitive. If there's no version at all, that doesn't mean the major version is 0.

Why not just return false immediately in `checkModuleSupport` if `version` is undefined?

I think ideally `majorVersion` has type signature `string => number` and Flow should catch errors if undefined is passed. Passing `undefined` to the `majorVersion` function makes no sense.

rtsao:
What is `fusion-cli/node/.block_boxer`? Should that be committed?

PR #566, opened 2019-07-23
Fix existing lock file corruptions, if any

UberOpenSourceBot:
The previous PR prevents yarn.lock corruptions from happening when generating a lockfile. This one fixes existing corruptions, if it detects any


AlexMSmithCA:
No-op.  Closing this PR.

PR #571, opened 2019-07-23
Add Promise<void> as valid callback return for universalEvents.on handler

raymondwang:
This trivially provides Flow support for an async publish call.


AlexMSmithCA:
This seems fine to me.  In theory, you can get around this by doing an asynchronous action while not awaiting it (and not returning a promise).

Just curious though, what is the concrete use case you are trying to meet here?

raymondwang:
> This seems fine to me. In theory, you can get around this by doing an asynchronous action while not awaiting it (and not returning a promise).
> 
> Just curious though, what is the concrete use case you are trying to meet here?

It probably isn't _strictly_ necessary, but we'd like to `await` for the catch block:

```
const publish = async payload => {
  try {
    await heatpipe.asyncPublish(HEATPIPE_TOPIC, payload);
  } catch (e) {
    logger.warn('analytics_plugin_error', e);
  }
};
// $FlowFixMe — this commit addresses this line:
universalEvents.on(WRM_ANALYTICS_EVENT_NAME, publish);
```

AlexMSmithCA:
!merge

PR #574, opened 2019-07-23
Update dependency versions in fusion-plugin-apollo

ganemone:


albertywu:
While you're at it, can you make `react-apollo` use the same version in `devDependencies` and `peerDependencies`?

ganemone:
!merge

PR #579, opened 2019-07-24
Fix promise instrumentation

UberOpenSourceBot:
There are 2 issues that are somewhat related, instrumented module ids and instrumented chunk ids.

### Issue - Module IDs
Some production builds were getting these errors:

`Bundle does not contain a default export ` from [here](https://github.com/fusionjs/fusionjs/blob/master/fusion-react/src/async/split.js#L108)

or

`Cannot read property 'call' of undefined` because `modules[moduleId].call(/* ... */)` in webpack bootstrap code.

### Cause
This seems to have been because the instrumented `__MODULE_ID` was different than the `__webpack_modules__` key.

### Fix
The initial implementation of the translation instrumentation, I was unsure why some production builds were missing a `module.id`, but it appears to be because `afterOptimizeDependencies` is an earlier compilation phase then `beforeModuleIds ` where the module ids are assigned in HashedModuleIdsPlugin, which is only used in production builds for client bundles.

This fix creates custom module ids at the `afterOptimizeDependencies` phase, and makes sure they are reused in the `beforeModuleIds` phase by caching them based on module path (from `module.libIdent`).

### Issue - Chunk IDs
In some places, translations were not working correctly. Serialized translations were empty or incomplete and `/_translations` API endpoint was responding with an empty object.

### Cause
This is caused by a disconnect between chunk ids on the server versus the client for the same module. Also, for `ContatenatedModules`, the instrumentation only included translations from the `originRequest` module.

### Fix
Use `module.rootModule.request` and `buildInfo.fileDependencies` to retrieve translations for `ConcatenatedModules`.

### DIFF
--- a/fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
+++ b/fusion-cli/build/plugins/instrumented-import-dependency-template-plugin.js
@@ -123,7 +123,6 @@ InstrumentedImportDependency.Template = class InstrumentedImportDependencyTempla
         'Dependency is not Instrumented and lacks a clientChunkIndex'
       );
     }
-
     const content = runtime.moduleNamespacePromise({
       block: dep.block,
       module: dep.module,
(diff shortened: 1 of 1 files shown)


AlexMSmithCA:
No-op.  Closing this.

PR #592, opened 2019-07-26
Use  'source-map' instead of 'hidden-source-map' for production builds

### LINKED ISSUE #349: [fusion-cli#763] Allow configuring 'source-map' instead of 'hidden-source-map' for production builds.
This issue was migrated from fusion-cli and was originally reported by @lxe.

---

Fusion-cli correctly `.map` files for production builds, but uses `hidden-source-map` webpack config, which omits the `//# sourceMappingURL` comment at the end of the source files.

From https://webpack.js.org/configuration/devtool/:

> hidden-source-map - Same as source-map, but doesn't add a reference comment to the bundle. Useful if you only want SourceMaps to map error stack traces from error reports, but don't want to expose your SourceMap for the browser development tools.

For many projects, it's useful to expose hidden source maps to devtools to allow debugging in production. 

I'd like to propose an option to use 'source-map' instead if 'hidden-source-map'

qqingvarqq:
Resolves #349
----
1. Use `source-map` for production build;
2. Use `none` for sw build since sw-loader doesn't produce the source map and changing from `hidden-source-map` to `none` increase build speed. More info: [webpack-devtool-options](https://webpack.js.org/configuration/devtool/#devtool);
3. I didn't remove `// TODO(#6): what about node v8 inspector?` since I don't know how you dealing with todos. (About this todo, It is possible to use chrome dev tools for debugging node.js code, which supports `cheap-module-source-map` best)


rtsao:
I think by default source map comments should not be included in production builds (it is assumed source map files will not be available to the client).

But for situations such as staging deploys, it may be useful to include the source map comments to make client debugging easier. In this case, I think it would make sense to have this configured via an environment variable.

micburks:
I agree that this shouldn't be the default. Perhaps a `.fusionrc.js` property would be sensible here for customization.

```js
module.exports = {
  sourceMapType: process.env.IS_STAGING_ENVIRONMENT && 'source-map'
  // falsy falling back to default
};
```

The name `sourceMapType` implies that you can choose a type of source map but a falsy value doesn't mean **no** source maps. Disabling source maps altogether could be another option, but conflating those 2 ideas would potentially cause confusion.

AlexMSmithCA:
Bump FYI @qqingvarqq

qqingvarqq:
Sorry for the delay, was very busy and didn't have much time for open source. Hope, I will update requested changes on this weekend.

mlmorg:
Closing for now. Please reopen when ready.

Issue #573, opened 2019-07-23
Unclear behavior for enhancers in certain edge cases


ganemone:
After investigating #572 I realized we have some undefined behavior surrounding certain edge cases for enhancers. The following edge cases have the following behavior:

**Enhancer on unregistered token with no dependencies**
  - does not throw at resolution time
  - does not execute enhancer
  
Example:
```js
  const app = new App();
  app.enhance(SomeEnhancer, TokenA);
  return app;
```

**Enhancer on unregistered token with optional dependency**
  - throws at resolution time

Example:
```js
  const app = new App();
  app.enhance(SomeEnhancer, TokenA);
  app.register(createPlugin({
    deps: {
      a: TokenA.optional
    }
  }));
  return app;
```

I think our strategy for DI logic should be focused on correctness, eliminating dead code, and validating intent as much as possible. This is why we have features such as disallowing orphan configuration registrations. The only use case I can think of for allowing enhancers to be registered on unregistered tokens would be for providing a default implementation. For example:

```js
app.enhance(TokenA, (a) => {
  if (!a) {
    return 'some-default';
  }
  return a;
});
```

However, I think mixing semantics of enhancing existing registrations and providing default implementations is not a good idea. If we want to support enhancing optional tokens, we can do so by extending the `app.enhance` semantics to support it:

```js
app.enhance(TokenA.optional, (a) => {
  if (!a) {
    return 'some-default';
  }
  return a;
});
```

Alternatively, we could introduce a new API specifically designed for providing default implementations (which we have discussed and rejected in the past).

```js
app.default(TokenA, 'some-default');
```

For now, my recommendation would be to fix #572 and to fail dependency resolution if an enhancer is registered on a Token with no implementation. If we have a need in the future for providing default implementations on Tokens, we can explore options listed above.

rtsao:
I think one drawback to the notion of "default" implementations is it leads to dead code if the default is replaced with something else. Are there specific use cases where this would be really helpful?

Obviously I could imagine it being more convenient, but this essentially encourages dead code while also increasing the complexity of understanding how/which values become registered at a specific token.

ganemone:
I agree about the dead code argument. I haven't seen any specific use cases where this would be needed yet, so I think it is fair to not implement anything surrounding defaults unless a compelling use case arises.

PR #593, opened 2019-07-26
Fix edge case with optional dependencies

### LINKED ISSUE #572: Dependency resolution allows invalid tree in edge case with optional dependencies
If a plugin is registered with an optional dependency on TokenA, and later a plugin is registered with a required dependency on TokenA, and nothing gets registered on TokenA, fusion-core will not fail the resolution and instead will provide `undefined` for the injected value for TokenA. For example:

```js
function Main() {
  const app = new App();
  app.register(createPlugin({
    deps: {
      a: TokenA.Optional
    },
    provides({a}) => {
      console.log(a); // undefined
    }
  }));
  app.register(createPlugin({
    deps: {
      a: TokenA
    },
    provides({a}) => {
      // Dependency resolution should fail but it does not.
      console.log(a); // undefined
    }
  }));
  return app;
}
```

ganemone:
Fixes #572 and adds some skipped regression tests for #573

### DIFF
--- a/fusion-core/src/__tests__/enhance.js
+++ b/fusion-core/src/__tests__/enhance.js
@@ -166,6 +166,94 @@ tape('enhancement with a plugin with deps', t => {
   app.resolve();
 });
 
+// TODO(#573) This is a regression test which hasn't been fixed yet
+tape.skip('Plugin enhancer on unregistered token with no dependencies', t => {
+  let app = new App('el', el => el);
+  type FnType = string => string;
+  const FnToken: Token<FnType> = createToken('FnType');
+  let BaseFnEnhancer = (fn: FnType): FusionPlugin<void, FnType> => {
+    return createPlugin({
+      provides: () => {
+        return arg => {
+          return fn(arg) + ' enhanced';
+        };
+      },
+    });
+  };
+  app.enhance(FnToken, BaseFnEnhancer);
+  t.throws(() => {
+    app.resolve();
+  });
+  t.end();
+});
+
+// TODO(#573) This is a regression test which hasn't been fixed yet
+tape.skip(
+  'Plugin enhancer on unregistered token with optional dependency',
+  t => {
+    let app = new App('el', el => el);
+    type FnType = string => string;
+    const FnToken: Token<FnType> = createToken('FnType');
+    let BaseFnEnhancer = (fn: FnType): FusionPlugin<void, FnType> => {
+      return createPlugin({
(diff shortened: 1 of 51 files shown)


fusion-core/src/base-app.js
```diff
@@ -1,5 +1,7 @@
 /** Copyright (c) 2018 Uber Technologies, Inc.
  *
+
```
AlexMSmithCA: Minor: remove new lines

fusion-core/src/__tests__/enhance.js
```diff
@@ -166,6 +166,94 @@ tape('enhancement with a plugin with deps', t => {
   app.resolve();
 });
 
+// TODO(#573) This is a regression test which hasn't been fixed yet
```
AlexMSmithCA: Thoughts on an additional plugin enhancer test where the enhancer's returned plugin has unregistered dependencies?  Something like:

```js
let app = new App('el', el => el);
type FnType = string => string;
const FnToken: Token<FnType> = createToken('FnType');
const EnhancerDepToken: Token<FnType> = createToken('EnhancerDepToken');
let BaseFnEnhancer = (fn: FnType): FusionPlugin<void, FnType> => {
  return createPlugin({
    deps: {
      addlFn: EnhancerDepToken,
    },
    provides: () => {
      return arg => {
        return fn(arg) + ' enhanced, and also ' + addlFn(arg);
      };
    },
  });
};
app.register(FnToken, msg => msg);
app.enhance(FnToken, BaseFnEnhancer);
t.throws(() => {
  app.resolve();
});
t.end();
```

(If we don't already cover this case)

ganemone: We do have this test case

fusion-core/src/base-app.js
```diff
@@ -172,73 +173,74 @@ class FusionApp {
...
-          this.register(token, undefined);
-        } else {
-          const dependents = Array.from(this.registered.entries());
+        // Early return if token is optional
+        const isOptional =
+          token instanceof TokenImpl && token.type === TokenType.Optional;
+        if (isOptional && (!enhancers || !enhancers.length)) {
+          return;
+        }
+        const dependents = Array.from(this.registered.entries());
```
AlexMSmithCA: Lines below here appear to be flagged for indentation as opposed to business logic changing.

ganemone:
!merge

PR #604, opened 2019-08-01
Fix typing of optional property in Token

phou87:
In https://github.com/fusionjs/fusionjs/pull/550, we changed the way we extract a Token's underlying service type, but failed to update the 'optional' field which was still using the old way. This PR fixes that by properly applying optionality (but not nullability - super important!) to the underlying type.

I tested this on a fairly large codebase and Flow seemed happy.

### DIFF
--- a/fusion-core/src/types.js
+++ b/fusion-core/src/types.js
@@ -10,7 +10,7 @@ import type {Context as KoaContext} from 'koa';
 
 export type Token<T> = {
   (): T,
-  optional: () => void | T,
+  optional: Token<void | T>,
   stacks: Array<{
     // eslint-disable-next-line
     type:


fusion-core/src/types.js
```diff
@@ -10,7 +10,7 @@ import type {Context as KoaContext} from 'koa';
 
 export type Token<T> = {
   (): T,
-  optional: () => void | T,
+  optional: Token<void | T>,
```
AlexMSmithCA: What are the implications here if we instead did `optional: ?Token<T>` which matches the `create-token.js` implementation more closely:

https://github.com/fusionjs/fusionjs/blob/73914f34f02187c54c9da51e5d7cc48e94aedf74/fusion-core/src/create-token.js#L16-L35

phou87: I actually think the `create-token.js` implementation is wrong, since it's optional but not nullable. If I try to make the type nullable instead, there will be a ton of other errors (example where a prior check would now fail: https://github.com/fusionjs/fusionjs/blob/master/fusion-plugin-styletron-react/src/browser.js#L42).

AlexMSmithCA: Good point.  Given that `?Token<T>` is equivalent to `null | void | Token<T>`, how about `optional: void | Token<T>` instead?

AlexMSmithCA: Discussed offline:

Larger issue is that the types in `types.js` and the implementation are inconsistent.  Changes that make `optional` potentially `null` or `void` cause regressions that we would need to fix.

Current proposal keeps us relatively close to the runtime behavior, but allows some odd behaviors (e.g. `token.optional.optional.optional...`).

phou87: Discussed offline - the north star solution here is a bit trickier to get to, whereas the current approach at least mirrors the old logic to avoid breaking other things.

Idea: make the `optional` field non-optional. Have the required token's `optional` point to the optional token, and have the optional token's `optional` point to itself. Someone more familiar with the codebase should probably play around with that, though.

AlexMSmithCA:
!merge

PR #608, opened 2019-08-02
Sync public<>private monorepo

AlexMSmithCA:
It looks like we somehow missed the force pushes in https://github.com/uber/fusionjs/pull/359 which weren't reflected in our mirrored PR: https://github.com/fusionjs/fusionjs/pull/600


AlexMSmithCA:
Going to break glass and land this.

PR #629, opened 2019-08-12
Add 'third_party' node_modules path to .flowconfig as part of flow check

UberOpenSourceBot:
Fixes [WPT-3920](https://jeng.uberinternal.com/browse/WPT-3920):

> By default, Flow will follow symlinks provided that they lead to a file or directory that is included.
> By default, only the root directory of the project (i.e. where .flowconfig is located) is included.
> Given that Jazelle points to <MONOREPO_ROOT>/third_party/jazelle/temp/node_modules, we'll want to add that to every .flowconfig.

### DIFF
--- a/common/config/rush/yarn.lock
+++ b/common/config/rush/yarn.lock
@@ -1529,7 +1529,7 @@
 
 "@rush-temp/create-fusion-app@file:./projects/create-fusion-app.tgz":
   version "0.0.0"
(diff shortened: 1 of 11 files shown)


AlexMSmithCA:
Oof, looks like `fusionjs-bot` auto-merged here with inconsistent changes.  I'll make a separate PR to handle that.

PR #642, opened 2019-08-19
Add failure RPC event when fetch fails

brendean:
Right now if the fetch fails (eg browser loses internet connection) no error event gets emitted, this pull adds a catch statement that will emit the error event and then rethrow the exception.


ganemone:
Looks like there are some auto-fixable lint errors.

PR #680, opened 2019-08-30
Use local package registry when installing packages and dependencies

derekju:
Currently, when a package is installing its dependencies and those dependencies are in different subtrees of the monorepo with different yarn registries, the update and sync methods in jazelle are not aware of the difference in registries when it stitches the lockfile back together.

During the stitching process, jazelle will index all of the lockfiles and then as it builds the lockfiles back up, it will take the highest versioned package across all deps and use that which is correct. However, if one of the projects was resolved with a different package registry, jazelle will potentially overwrite it indiscriminately since it has no concept of ensuring that the registry is correct for that package.

This fix uses liberal use of yarn config get registry before installing missing and transitive deps so that the correct registry for the package will be used.

In the case of lockfile syncing, this fix will convert all lockfile resolved paths to a relative path instead to be used as the index. Then when lockfiles are rebuilt, the registry that is local to the package (again using yarn config get registry) will be prepended to the relative path to reconstruct the lockfile.

There is an open issue in the yarn repo that discusses changing the paths for good to relative paths but there seems to be no recent movement on this: yarnpkg/yarn#5892

The end result of this PR is that:

multiple projects in the monorepo can have their own .yarnrc files that point to different registries
jazelle will ensure all dependencies are synced across all projects it controls while respecting the configured registry for each project


chrisdothtml:
for reference, this is a remake of https://github.com/fusionjs/fusionjs/pull/672

derekju:
!import

PR #684, opened 2019-09-09
Add FormData support into fusion-plugin-rpc

ardok:


ganemone:
Does this work with the default body parser options?

ardok:
@ganemone are you talking about the BodyParserOptionsToken from `fusion-plugin-rpc-redux-react`?

Either way, I'm not sure. I assumed that the existing tests should have covered some basic cases for that?
If not, is there any test example in other plugins using this "default body parser"?

ganemone:
I'm not sure if we have an existing test since it would need to be an integration test of sorts, but you could test it out in an application.

ardok:
Oh yeah, that works just fine.
I changed the `fusion-plugin-rpc-redux-react/dist/node_modules/fusion-plugin-rpc/dist/browser.es2017.es.js` code to this:
```
    return fetch(
      `/api/${rpcId}`,
      args instanceof FormData
        ? {
          method: 'POST',
          headers: {
            // Content-Type will be set automatically
            ...(headers || {}),
          },
          body: args,
        }
        : {
          method: 'POST',
          headers: {
            'Content-Type': 'application/json',
            ...(headers || {}),
          },
          body: JSON.stringify(args || {}),
        }
    ).then(...)
```

And my app still works fine (including the file upload).

Is that sufficient though?

ganemone:
Sounds good. Looks like you have some formatting that needs fixing, and also should rebase

ardok:
Can someone re-add the ci label or whatever it is to re-run the tests?

ardok:
Lol. This is annoying. I just merged with master and it needs that ci label again to be added

ganemone:
!land

ganemone:
!import

PR #691, opened 2019-09-24
Issue #548 Add an ability to register multiple redux stores in a single application

romaxa:
Add an ability to register multiple redux stores in a single application

ReducerNameSpaceToken allows separating redux store cache by the unique key.
This allows creating multiple redux stores per application instance, as well as multiple fusion apps with different redux stores


chrisdothtml:
Closing in preference of PR in private repo

PR #693, opened 2019-10-01
Ensure translation keys are an array

brendean:
Previously we were using the body if present but body will be an empty object after being parsed. Added more checks to ensure we marshall the keys into an array otherwise server crashes could be caused by sending bad requests.


chrisdothtml:
!import

PR #697, opened 2019-10-22
Allow error-handling plugin to accept transform function [fusion-plugin-error-handling] feature(UBERCOM-2530)

foodforarabbit:
resolves https://github.com/fusionjs/fusionjs/issues/698

### DIFF
--- a/fusion-plugin-error-handling/src/__tests__/index.node.js
+++ b/fusion-plugin-error-handling/src/__tests__/index.node.js
@@ -7,14 +7,16 @@
  */
 
 /* eslint-env node */
-
 import test from 'tape-cup';
 import {fork} from 'child_process';
 
 import App, {consumeSanitizedHTML} from 'fusion-core';
 import {getSimulator} from 'fusion-test-utils';
 
-import ErrorHandling, {ErrorHandlerToken} from '../server';
+import ErrorHandling, {
+  ErrorHandlerToken,
+  ErrorHandlingTransformToken,
+} from '../server';
 
 test('request errors', async t => {
   t.plan(6);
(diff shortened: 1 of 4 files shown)


micburks:
Hi! Thank for the PR. I've got some feedback.

I'm not sure serializing a function and adding all this logic is necessary for what should be a
fairly straightforward error handler. I think it'd be better to apply any logic you need to the `/_errors` endpoint by overriding or enhancing the `ErrorHandlerToken`. FWIW, all the information you mentioned in the linked issue should be readily available on the koa context, `ctx.request.header.referer`, `ctx.request.header.cookie`, etc.

foodforarabbit:
I've tried implementing it purely using the ErrorHandlerToken (by modifying the @uber/fusion-plugin-error-handling package) but unfortunately, the ctx available to the error handler is the ctx for the request to `/_errors`, and no information available from the original request besides what gets `post`ed by the client in the onerror function. The problem is that any errors coming from the frontend is another request all together, so we don't have access to any information about the request that is triggering errors in the first place, unless we do something like cache request data with a key so we can find information about the original request. I agree that serializing a function is kinda ugly though - especially since it wont be going through the proper webpacking channels (the function must be self-contained with no imports).

Can you advise me on an implementation path?

micburks:
> but unfortunately, the ctx available to the error handler is the ctx for the request to `/_errors`...

This is correct, but the `referer` of the request is the page from which the `/_errors` request is being made. For example, the `ctx.request.header.referer` for a `/_errors` request made from the `www.mysite.com/some/path` is `www.mysite.com/some/path`. In addition, the cookies are sent with this request, so you can access all that.

> Can you advise me on an implementation path?

```js
import {createPlugin} from 'fusion-core';
import ErrorHandlingPlugin, {
  ErrorHandlerToken,
} from 'fusion-plugin-error-handling';

export default function (app) {

  // Mock some browser error
  if (__BROWSER__) {
    setTimeout(() => {
      throw new Error('Log this error!');
    }, 1000);
  }

  if (__NODE__) {
    app.register(ErrorHandlerToken, MyErrorHandlingPlugin);
    app.enhance(ErrorHandlerToken, baseHandler => {
      // [How to enhance](https://fusionjs.com/api/fusion-core#enhancing-a-dependency)
      return createPlugin({
        provides() {
          return function (body, captureType, ctx) {
            // Do whatever you want here
            console.log(ctx.request.header.referer);
            console.log(ctx.request.header.cookie);

            // But still pass off to the plugin you are enhancing
            return baseHandler(body, something, ctx);
          };
        }
      });
    });
}
```

foodforarabbit:
@micburks this works for what we need, thanks!

PR #704, opened 2019-11-13
Fix: FRAMEWORK_STATIC_ASSETS_PATH not being used in client side

Monar:
Currently FRAMEWORK_STATIC_ASSETS_PATH was not utilized in all places. 
If set the server side would server assets under correct path but client side code would ignore this variable and fetch from default `/_static` code path. 

This change fixes this problem by setting proper `assetsBasePath` utilizing FRAMEWORK_STATIC_ASSETS_PATH variable.


Monar:
So, ci fails due to eslint getting excessive parameter, eg.:

```

Done in 49.94s.
--
  | yarn run v1.16.0
  | $ eslint . --cwd=fusion-plugin-service-worker
  | Invalid option '--cwd' - perhaps you meant '-c'?

``` 

So as I understand I just wait for the fix, yes ?

chrisdothtml:
> So, ci fails due to eslint getting excessive parameter

@Monar sorry for the delay! We had some issues in our CI logic that have now been resolved

ganemone (approved):

rtsao:
FWIW I think `FRAMEWORK_STATIC_ASSET_PATH` is kind of undocumented in favor of using either `CDN_URL` or `ROUTE_PREFIX`. I think this still exists only for legacy reasons.

I wonder if we should just remove this since I don't think `/_static` is meant to be configurable.

rtsao:
Yeah, there's a few of spots where `/_static` is hardcoded, so I don't think even with this change this will work properly

KevinGrandon (changes requested):
Going to request changes based on rtsao's latest comment. Seems like this shouldn't be configurable. Is it causing problems somehow?

PR #705, opened 2019-11-14
Fix this.translations === null in translate function

zirho:
I encountered this === null type error in one of the service.

```
TypeError: Cannot read property 'app.title' of undefined\n    at I18n.translate (/home/udocker/safety-lens/node_modules/fusion-plugin-i18n/src/node.js:47:28
```

This happens in SSR

The app was migrated from bedrock and don't know if my config messed it up but wanted to prevent this from happening even with the wrong config

This happens when we don't provide any other language translations other than English and users set browser locale to different languages.

I was expecting it to fallback to English automatically but it seems it's not the case.

### DIFF
--- a/fusion-plugin-i18n/src/node.js
+++ b/fusion-plugin-i18n/src/node.js
@@ -105,7 +105,7 @@ const pluginFactory: () => PluginType = () =>
         }
         async load() {} //mirror client API
         translate(key, interpolations = {}) {
-          const template = this.translations[key];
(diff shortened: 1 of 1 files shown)


zirho:
I don't know if this is specific to Spanish locale but this is how it looks on the computer that exhibit this error.

Image Pasted at 2019-11-13 21-06

zirho:
closing it because I confirmed that my app config was wrong

PR #706, opened 2019-11-19
Feature: add option to limit number of workers spawn during build

Monar:
When building fusion application in docker, number of spawn workers is relative to host machine available "cpus" and not a number of cpu assigned to docker image. This leads to failed build and only arbitrary/explicit limit of max concurrent process can solve this issue. 

This PR is introducing `--maxWorkers` parameter to build command that limit default number equal to `os.cpus.length - 1`


chrisdothtml:
!import

PR #707, opened 2019-11-20
Avoid frequent require to "package.json"

ravihamsa:
`mapPerfEvent` will get called once per server request
having `require('../package.json')` would mean reading and parsing package.json
happens once per request. This change will help stopping those frequent parsing


ravihamsa:
@ganemone looks like this stuck, can you please help me to unblock it?

chrisdothtml:
!import

chrisdothtml:
@ravihamsa sorry for the big delay. We had some issues in our CI logic that have now been resolved

PR #710, opened 2019-12-04
Change Logger flow definition

cgvedant:
Type definition of `Logger` was recursively using `Logger`. https://github.com/uber/logtron suggests `void` is the correct return type.


derekju:
@cgvedant: Looks like you have some test failures, can  you fix them locally first?

cgvedant:
> @cgvedant: Looks like you have some test failures, can you fix them locally first?

done

angus-c:
I think the old code is correct as stands. Our plugin interface expects `log` methods to return a logger (doesn't follow Logtron).
https://github.com/uber/fusionjs/blob/master/private/fusion-plugin-logtron/src/server.js#L85

PR #716, opened 2020-01-10
Add dot-notation error rule

ajbogh:
The dot-notation ensures variable access uses dot-notation rather than array notation where array accessors would be unnecessary. Numeric array indices, variable indices, and template literals are still accessed with array notation as normal, only static strings would be changed.

Using the --fix option automatically fixes all variables.

Bad: 

```
const myVar = SomeObject.someProp["someKey"].otherProp[0];
```

Good: 

```
const myVar = SomeObject.someProp.someKey.otherProp[0];
```


chrisdothtml:
!import

PR #717, opened 2020-01-21
Add Shebang

DylanVann:
I'm using fish as my shell. Currently when trying to use Jazelle I get:

image

This PR resolves this by specifying that these scripts should use Bash.


DylanVann:
Maybe I'm doing something wrong but even after this fix I don't see `jazelle install` generating any `BUILD.bazel` files.

lhorie:
!import

PR #719, opened 2020-01-25
Update README.md

ajbogh:
Based on the shareable configs doc we don't need resolve.require to include fusion linting rules. We also don't need "eslint-config-" because the system assumes that.

"You can also omit the eslint-config- and it will be automatically assumed by ESLint:"

https://eslint.org/docs/developer-guide/shareable-configs


chrisdothtml:
Hey, thanks for the PR. Personally this seems like a stylistic preference more than anything. For example, the current method is more explicit and obvious to someone not already familiar with the eslint ecosystem.

Closing this unless it's actually causing any issues

PR #720, opened 2020-01-25
Update index.js

ajbogh:
Flowtype rules shouldn't be included in the Fusion eslint rules because Fusion can be run on systems that don't include flowtype, such as Typescript. Requiring these rules would force projects to install eslint-plugin-flowtype even though they don't need it to run Fusion. Those projects would then have to include the following to disable flowtype linting errors:

```
rules: {
    "flowtype/require-valid-file-annotation": 0,
    "flowtype/no-types-missing-file-annotation": 0,
```

A recommendation is to split flowtype and typescript rules into their own packages, such as eslint-config-fusion-flowtype and eslint-config-fusion-typescript. The users of Fusion can choose which package to install, or the scaffolding tool can select the appropriate package depending on the type of project (if that choice becomes available in the future).

### DIFF
--- a/eslint-config-fusion/index.js
+++ b/eslint-config-fusion/index.js
@@ -1,7 +1,6 @@
 // @flow
 module.exports = {
   extends: [
-    'plugin:flowtype/recommended',
     'plugin:react/recommended',
     'plugin:jest/recommended',
     './rules/imports.js',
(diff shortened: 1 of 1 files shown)


chrisgervang:
This proposal makes sense and follows fusions style of splitting components into smaller libraries.

AlexMSmithCA:
:+1: This approach sounds reasonable to me.

In order to support folks using the Flow rules today, we would also (1) need to publish the Flow-specific ruleset, and (2) need to include a codemod to add a dependency on the Flow-specific package in [create-uber-web](https://github.com/uber/fusionjs/tree/master/private/create-uber-web) (internal link).

chrisdothtml:
@ajbogh I would recommend remaking this PR in the private fusion monorepo to include all the changes @AlexMSmithCA mentioned. Feel free to ping me if you need any help

ajbogh:
@chrisdothtml I'll be glad to attempt the requested codemod for the create-uber-web project. I'll add 2 additional public PRs for the new packages so that this work can be tracked in the public repo and conversations can be available to everyone.

ajbogh:
This should not be merged until should #723 is approved and merged first AND a codemod is available for internal projects. 

It's recommended that eslint-config-fusion use a major version bump when this is merged.

gutenye:
#723 has been merged, any updates on this one? (I'm trying to use fusionjs with typescript and get blocked by this).

PR #721, opened 2020-01-27
Update flowtype for prepared

akre54:
`sideEffect` doesn't actually need to return a Promise since it is wrapped once in Promise.resolve and the other two times the return type is discarded.

### DIFF
--- a/fusion-react/src/async/prepared.js
+++ b/fusion-react/src/async/prepared.js
@@ -19,7 +19,7 @@ type PreparedOpts = {
 };
 
 const prepared = (
-  sideEffect: (any, any) => Promise<any>,
+  sideEffect: (any, any) => any | Promise<any>,
   opts?: PreparedOpts = {}
 ) => <Config>(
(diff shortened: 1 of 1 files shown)


akre54:
`prepareState.isResolved` pushes the `effectPromise` into a Set which is [eventually consumed by `Promise.all`](https://github.com/fusionjs/fusionjs/blob/cc57962a8ba21b7b07936b93a2a0f4c26735d769/fusion-react/src/async/prepare.js#L63).

Promise.all casts `null` to a resolved promise.
```
> p = Promise.all([null, Promise.resolve(3)])
Promise {<resolved>: Array(2)}
__proto__: Promise
[[PromiseStatus]]: "resolved"
[[PromiseValue]]: Array(2)
0: null
1: 3
length: 2
__proto__: Array(0)
> p = await Promise.all([null, Promise.resolve(3)])
(2) [null, 3]

```

chrisdothtml:
!import

PR #724, opened 2020-01-28
Add collectCoverage as an allowed Jest option

ajbogh:
I believe "coverage" should have been "collectCoverage" since Jest lists the option as "collectCoverage". There is no "coverage" option in Jest. In order to keep in line with the valid Jest options, we should accept "collectCoverage".

I'm not entirely sure what "coverage" does, since a boolean true value for it in the .fusionrc.js file doesn't change anything when `yarn test` is run.

https://jestjs.io/docs/en/configuration#collectcoverage-boolean

This PR also adds coverageThreshold as an allowed option. See the following commit:

https://github.com/fusionjs/fusionjs/pull/724/commits/9000000b8ffe40b883e50c6365a9e1a5b81593dc


chrisdothtml (dismissed):

chrisdothtml:
@ajbogh can you update the README?

ajbogh:
@chrisdothtml I added a note about the collectCoverage option. The coverageThreshold is only a Jest parameter that's used in the config files. I don't think this should be added to the README, but I believe not having the option in the cli-options file is preventing it from being used.

chrisdothtml:
> I believe "coverage" should have been "collectCoverage" since Jest lists the option as "collectCoverage". There is no "coverage" option in Jest.

Hmm seems the initial description of your pr is inaccurate based on the docs. I do see [a `--coverage` arg](https://jestjs.io/docs/en/cli.html#--coverageboolean) documented, with `--collectCoverage` listed as an alias (even in [the v24 docs](https://jestjs.io/docs/en/24.x/cli#--coverage), which is the version `fusion-cli` is currently on). I think it's fine to pass through the alias (`--collectCoverage`), but now I'm more interested in why passing `--coverage` isn't doing anything for you.

Are you trying this with latest `fusion-cli`?

ajbogh:
@chrisdothtml Thanks for the help.

It is the latest fusion and fusion-cli. Maybe you can help me understand this. I've added Jest options to the .fusionrc file but some of those options are ignored. I had tracked it down to this object which is used to filter out accepted parameters.

The use case is I would like to specify which files and folders to collect coverage from, to enable coverage for `yarn test`, and provide various other options (doesn't matter what option, anything really).

The scaffold creates a Jest section in the .fusionrc file, but it seemed as though this configuration wouldn't allow certain parameters, such as the "coverage" option.

The question is, should I continue with this change and find out why these parameters in the .fusionrc are being ignored, or should we instead add a jest configuration to the package.json file, thereby overriding what Fusion might otherwise filter out.

ganemone (approved):

PR #731, opened 2020-02-06
Add `no-console` eslint rule

ericsoco:
I imagine There Will Be Discussion about the value of adding / not adding this rule. IMO, it's a valuable safeguard against committing debugging statements, and is easy to disable line-by-line for necessary `console` usage.

Also, it seems that [`eslint-config-uber-base`](https://github.com/uber-web/uber-eslint/blob/master/packages/eslint-config-uber-base/index.js) might be a better place for this, but @rtsao suggested I start with a PR to Fusion.


ericsoco:
Any thoughts on adding this rule?

PR #734, opened 2020-02-28
feat(events): make endpoint customizable

lessenadam:
Based on [GraphQLEndpointToken](https://github.com/fusionjs/fusionjs/tree/6e877e514f8b7e61c8a501a35a636e9fc12b8bf4/fusion-plugin-apollo#graphqlendpointtoken).

The server tests pass for both cases: when the token is provided, and when the token is omitted.

The browser tests pass when the token is omitted (ie, does not break functionality). There is some setup leak between tests, as the customized endpoint test passes when it's run as the only test, but somehow fails due to using an instance of the class from a previous test without the customized endpoint. Perhaps someone who is more familiar with this older test setup can take a look? Or perhaps it's test coverage enough that omitting the token **does not break** the existing functionality.

### DIFF
--- a/fusion-plugin-universal-events/__tests__/test.browser.js
+++ b/fusion-plugin-universal-events/__tests__/test.browser.js
@@ -16,7 +16,9 @@ import type {Fetch} from 'fusion-tokens';
 import {getSimulator} from 'fusion-test-utils';
 
 import plugin, {UniversalEmitter} from '../src/browser.js';
-import {UniversalEventsToken, UniversalEventsEndpointToken} from '../src/index';
+import {
+  UniversalEventsToken /* UniversalEventsEndpointToken */,
+} from '../src/index';
 import {
   UniversalEventsBatchStorageToken,
   inMemoryBatchStorage as store,
@@ -188,9 +190,9 @@ test('Lowers limit for 413 errors', async () => {
   await emitter.flush();
 });
 
-// this test passes if run as test.only
+// this test passes if run as test.only. need to uncomment the import statement
 
-// test('Browser EventEmitter URL is configurable', async () => {
+// test.only('Browser EventEmitter URL is configurable', async () => {
 //   let fetched = false;
 //   let emitted = false;
 //   const fetch: Fetch = (url, options) => {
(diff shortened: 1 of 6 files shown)


fusion-plugin-universal-events/__tests__/test.browser.js
```diff
@@ -187,3 +189,59 @@ test('Lowers limit for 413 errors', async () => {
   expect(emitter.limit).toBe(5);
   await emitter.flush();
 });
+
+// this test passes if run as test.only. need to uncomment the import statement
+
+// test.only('Browser EventEmitter URL is configurable', async () => {
```
ganemone: Whats going on with this test?

lessenadam: Would you be able to take a look? 

> There is some setup leak between tests, as the customized endpoint test passes when it's run as the only test, but somehow fails due to using an instance of the class from a previous test without the customized endpoint.

mlmorg:
Why would we make this configurable versus simply using the `ROUTE_PREFIX` configuration. It sounds like the issue you are having is because it is always hitting the `/` route which is a different application then yours. Using `ROUTE_PREFIX` would solve this and not require every app to set this configuration up as well as configure `ROUTE_PREFIX`...

lessenadam:
@mlmorg 

> Why would we make this configurable versus simply using the `ROUTE_PREFIX` configuration. It sounds like the issue you are having is because it is always hitting the `/` route which is a different application then yours. Using `ROUTE_PREFIX` would solve this and not require every app to set this configuration up as well as configure `ROUTE_PREFIX`...

I actually got it working with process.env.ROUTE_PREFIX. However, I saw a bigger issue in that the route_prefix found it’s way into the react-router. So every `<Link />` was being directed to “/pre-fix/foo” instead of “foo” which won’t work for my case. I'd also worry about potential other edge cases like that. Happy to explore that option if we can handle the unwanted side affects.

mlmorg:
Maybe I'm confused. Is your app configured with a `ROUTE_PREFIX` currently or not?

lessenadam:
The `ROUTE_PREFIX` was an experiment, but didn't work in this case since we don't have control over changing the actual routes where the actual application needed to be served. After talking with Giancarlo, I'm just going to deploy the service at its own domain as opposed to sharing an existing domain with other services.

PR #736, opened 2020-04-09
Add fusionrc option to override main entry point

pheew:
This will allow implementers to use a custom entry point. This opens up the ability to use FusionJS with Typescript or any other language that has babel support. I wanted to get some early feedback before I

- [ ] Add documentation
- [ ] Write tests


fusion-cli/build/get-webpack-config.js
```diff
@@ -132,9 +132,14 @@ function getWebpackConfig(opts /*: WebpackConfigOpts */) {
     legacyPkgConfig = {},
     worker,
   } = opts;
-  const main = 'src/main.js';
+  const main = fusionConfig.main || 'src/main.js';
```
ganemone, line 135: Could we instead just check for the existence of `src/main.js` or `src/main.ts` and update our webpack accordingly?

PR #738, opened 2020-04-18
WIP: refactor fusion-plugin-react-router to make depending on browser history easier

btford:
Hello!

I'm creating this PR to follow up on a conversation I had with @ganemone.

tl;dr– I'd like to be able to access browser history in a provider of one of my plugins. Being able to access history in a provider saves me from writing a lot of boilerplate downstream, which greatly simplifies my application and plugin code.

I'm curious if the approach below is something that makes sense– I pulled apart the router plugin into distinct server and browser plugins, and then I have just the browser plugin depend on the router history.

Thanks in advance!


fusion-plugin-react-router/README.md
```diff
@@ -174,6 +174,23 @@ app.register(GetStaticContextToken, (ctx: Context) => {
 });
 ```
 
+##### `BrowserHistoryToken`
```
btford, line 177: @ganemone – I made my best efforts here in terms of the docs. Let me know if you have feedback.

fusion-plugin-react-router/__tests__/plugin.js
```diff
@@ -412,6 +414,51 @@ if (__BROWSER__) {
...
+        const {history} = router.from(ctx);
+        expect(history).toBeTruthy();
+        middlewareHistory = history;
+        return next();
+      }
+    );
+    const simulator = setup(app);
+    await simulator.render('/');
+
+    expect(middlewareHistory).toBe(providerHistory);
```
btford, line 458: I mostly just want to assert that the instance is the same regardless of where you get it.

Is there anything else I should add to the tests?

ganemone: this is good

btford:
Thanks @ganemone, @lhorie!

lhorie:
There appears to be a lint error, but other than that, LGTM

btford:
The lint error appears to be:

```

 
--
  | /monorepo/fusion-plugin-react-router/src/plugin.js
  | 70:45  error  'window' is not defined  cup/no-undef
  |  


```

...but this code was present before my change.

What is the recommended way to get ahold of `window` here? Or should I ignore/remove the warning?

ganemone:
> The lint error appears to be:
> 
> ```
> 
>  
> --
>   | /monorepo/fusion-plugin-react-router/src/plugin.js
>   | 70:45  error  'window' is not defined  cup/no-undef
>   |  
> ```
> 
> ...but this code was present before my change.
> 
> What is the recommended way to get ahold of `window` here? Or should I ignore/remove the warning?

So the cup/no-undef rule is smart enough to recognize code that exists inside `__NODE__` and `__BROWSER__` blocks. On master, the usage of `window` was inside a `if (__BROWSER__)` block.

ganemone:
> Fixed the lint. Does buildkite need to be manually restarted?

Are you sure you pushed your changes? It doesn't look like anything has been updated

PR #740, opened 2020-04-21
Refactor fusion-plugin-react-router to make depending on browser history easier

chrisdothtml:
Remaking fusionjs/fusionjs#738 which was improperly merged

---

Hello!

I'm creating this PR to follow up on a conversation I had with @ganemone.

tl;dr– I'd like to be able to access browser history in a provider of one of my plugins. Being able to access history in a provider saves me from writing a lot of boilerplate downstream, which greatly simplifies my application and plugin code.

I'm curious if the approach below is something that makes sense– ~~I pulled apart the router plugin into distinct server and browser plugins, and then I have just the browser plugin depend on the router history.~~

Thanks in advance!

### DIFF
--- a/fusion-plugin-react-router/README.md
+++ b/fusion-plugin-react-router/README.md
@@ -174,6 +174,23 @@ app.register(GetStaticContextToken, (ctx: Context) => {
 });
 ```
 
+##### `BrowserHistoryToken`
+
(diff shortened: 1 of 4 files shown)


chrisdothtml:
!import

PR #741, opened 2020-04-21
Revert "Revert "Refactor fusion-plugin-react-router to make depending on browser history easier""

ganemone:
Reverts fusionjs/fusionjs#739


ganemone:
!land

ganemone:
!import

Issue #745, opened 2020-04-29
[fusion-plugin-react-router] uncaught URIError in createLocation


foodforarabbit:
The change made on https://github.com/fusionjs/fusionjs/commit/1b86afd3a3a06ec8bb4bc58d4c6b3e161e101acc#diff-95e59cd7cab57bd863c7419d4464fa98R24 introduced a regression.

history.createLocation attempts to decodeURI the pathname (https://github.com/ReactTraining/history/blob/master/modules/LocationUtils.js#L36) whereas the previous implementation would not. This will throw a URIError on invalid paths (e.g. /%C0%AE%C0%AE/%C0%AE%C0%AE/WEB-INF/web.xml/ which seems to be some kind of directory traversal attack.) and cause the server to 500.

The previous behaviour would just leave the path as-is, which usually just resulted in a 404.

PR #746, opened 2020-04-29
Fix issue #745 [fusion-plugin-react-router] uncaught URIError

foodforarabbit:
fixes issue fusionjs/fusionjs#745 

in ServerHistory, we currently use history.createLocation which attempts to decodeURI the pathname

this throws errors when given a pathname with invalid uri encoding such as /%C0%AE%C0%AE/

previously, we were using parsePath without the decoding so this seems like a regression

the fix is to catch the invalid decoding error and simply encode the pathname beforehand, nullifying the decodeURI step.

### DIFF
--- a/fusion-plugin-react-router/__tests__/plugin.js
+++ b/fusion-plugin-react-router/__tests__/plugin.js
@@ -139,6 +139,25 @@ if (__NODE__) {
     expect(ctx.res.getHeader('Location')).toBe('/test/lol');
     cleanup();
   });
+  test('handles url with invalid URI encoding', async () => {
+    const Hello = () => <div>Hello</div>;
+    const element = (
+      <div>
+        <Route path="/" component={Hello} />
+      </div>
+    );
+    const app = getApp(element);
+    const UniversalEvents = getMockEvents({
+      title: 'no-matching-route',
+      page: '/%C0%AE%C0%AE/',
+    });
+    app.register(UniversalEventsToken, UniversalEvents);
+    const simulator = setup(app);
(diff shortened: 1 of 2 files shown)


fusion-plugin-react-router/__tests__/plugin.js
```diff
@@ -213,7 +232,7 @@ test('Custom Provider', async () => {
 
   const app = getApp(element);
   const UniversalEvents = getMockEvents({
-    title: '/',
+    title: 'no-matching-route',
```
foodforarabbit, line 235: this was actually throwing an error, but the `expect` was async so it was erroring after the tests passed and completed.

fusion-plugin-react-router/__tests__/plugin.js
```diff
@@ -439,6 +458,29 @@ if (__BROWSER__) {
     const simulator = setup(app);
     await simulator.render('/');
   });
+  test('handles url with invalid URI encoding', async () => {
```
angus-c, line 461: is this a duplicate test?

foodforarabbit: one is for server and one is for browser

foodforarabbit: will rename them!

chrisdothtml:
@foodforarabbit looks like tests are failing for this

foodforarabbit:
> @foodforarabbit looks like tests are failing for this

looking into it

chrisdothtml:
!import

PR #753, opened 2020-06-14
Finish flush when we return, when there are no items to push

nm123github:
If we don't have any items to push the condition is triggered and we return. If we don't call finishFlush() looks like flushInternal would always be blocked.


nm123github:
Added a test to verify items are emitted, even after flush() is called with no items.

@ganemone @lhorie

lhorie:
!land

lhorie:
!import

PR #411, opened 2019-05-22
Add in fusionConfig: jsExtPattern, resolveExtensions  (enable typescript)

Monar:
This is change proposed in here: https://github.com/fusionjs/fusion-cli/pull/705

We are using typescript with fusionjs for last 3 month and its working smoothly, this is minimal (generic) change that allows to embrace typescript for the app code.

PR #759, opened 2020-10-22
feat: allow .ts extension and support main.ts

pheew:
This is an alternative to #411 for supporting Typescript. This is the 'batteries included' approach where `.ts` and `.tsx` extension are supported by fusion out of the box. 

It still leaves the actual type checking up to the consuming project (like it should).


sansdesign-albert:
What needs to happen to get this merged?

PR #760, opened 2020-11-05
Update fusion-plugin-csrf-protection plugin to include the route prefix for non-csrf method requests

hari-sr:
Add the route prefix to URL for non-csrf method requests like GET requests


derekju (approved):

lhorie (approved):

derekju:
!import

PR #765, opened 2021-04-30
Adding  hot reload short circuit for CSR apps

prateekbh:
- Adding a short circuit for  CSR apps that do not always touch their server-side bundle on dev time.
- This saves a substantial time of respawning server for mid to large-size applications.


fusion-cli/build/dev-runtime.js
```diff
@@ -140,6 +142,21 @@ module.exports.DevelopmentRuntime = function(
       }
     `;
 
+    /**
+     * An escape hatch to not respawn the server if the server bundle never changed.
+     * This saves substantial times for CSR apps, who rarely change code in their server bundle.
+     */
+    if (existsSync(entryFile)) {
```
slonoed, line 149: Nit. Good practice is to avoid using sync file operations. util/promisify can be used to wrap callback based functions.

